Issue #2, opened 2024-12-15
Setup Publishing and building with tsup


seveibar:
We want to add a build command and a `"bin"` binary for this module so it can be installed with `npm install -g @tscircuit/snippets-cli`

We should also use tscircuit plop to add a pver release

/bounty $20

PR #4, opened 2024-12-15
build with tsup-node

ShiboSoftwareDev:
/claim #2
I used tsup-node as tsup had too many problems, is that acceptable?
gh actions are not running


ShiboSoftwareDev:
@seveibar

seveibar:
@ShiboSoftwareDev completely by accident i robbed you on this one.... #6 

you did this completely correctly so it's a bit lame- sorry about that /tip $10

PR #32, opened 2025-01-20
feat: automatically refresh import types on update

### LINKED ISSUE #28: import types on server refresh
if there are import changes import the new types
currently you only import on server start

Image

ArnavK-09:
/claim #28

Fixes #28

### DIFF
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -8,7 +8,7 @@ import path from "node:path"
 import fs from "node:fs"
 import type { FileUpdatedEvent } from "lib/file-server/FileServerEvent"
 import * as chokidar from "chokidar"
-import { isFileImportsTyped } from "lib/dependency-analysis/isFileImportsTyped"
+import { isFileTypesImported } from "lib/dependency-analysis/isFileTypesImported"
 import { installNodeModuleTypesForSnippet } from "lib/dependency-analysis/installNodeModuleTypesForSnippet"
 
 export class DevServer {
@@ -129,7 +129,7 @@ circuit.add(<MyCircuit />)
     // because it can be edited by the browser
     if (relativeFilePath.includes("manual-edits.json")) return
 
-    if (!isFileImportsTyped(absoluteFilePath, this.resolvedTypedImports)) {
+    if (!isFileTypesImported(absoluteFilePath, this.resolvedTypedImports)) {
       console.log("Types refreshing...")
       const updatedImports =
         await installNodeModuleTypesForSnippet(absoluteFilePath)
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -115,11 +115,11 @@ circuit.add(<MyCircuit />)
 
   async handleFileChangedOnFilesystem(absoluteFilePath: string) {
     const relativeFilePath = path.relative(this.projectDir, absoluteFilePath)
-
     // We've temporarily disabled upserting manual edits from filesystem changes
     // because it can be edited by the browser
     if (relativeFilePath.includes("manual-edits.json")) return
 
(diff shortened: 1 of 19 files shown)


ArnavK-09:
Status: draft 
Todo typecheck

ArnavK-09:
Draft

cli/dev/DevServer.ts
```diff
@@ -8,6 +8,8 @@ import path from "node:path"
 import fs from "node:fs"
 import type { FileUpdatedEvent } from "lib/file-server/FileServerEvent"
 import * as chokidar from "chokidar"
+import { isFileImportsTyped } from "lib/dependency-analysis/isFileImportsTyped"
```
imrishabh18: This name is a bit confusing to me, what do you think of this?
```suggestion
import { isFileTypesImported } from "lib/dependency-analysis/isFileTypesImported"
```

ArnavK-09: update, can now it be merged?

thanks

imrishabh18:
Let's see what @seveibar has to say on this.

seveibar (changes requested):
The DevServer should have a "user-friendly" interface, it probably should handle the type importing etc inside of it.

This is a class people would import to start the DevServer in eg a script, but this refactor would make it so it was dependent on code for type management

ArnavK-09:
> The DevServer should have a "user-friendly" interface, it probably should handle the type importing etc inside of it.
> 
> This is a class people would import to start the DevServer in eg a script, but this refactor would make it so it was dependent on code for type management


I’ve updated the code, and this represents the optimal solution I can provide at this time. I hope you find it useful. Kindly review it and suggest any changes so we can proceed to merge. Thanks!

image

cli/dev/DevServer.ts
```diff
@@ -155,4 +168,47 @@ circuit.add(<MyCircuit />)
...
+
+    const typePath = path.join(
+      projectRoot,
+      "node_modules",
+      "@tsci",
+      `${owner}.${name}`,
+      "index.d.ts",
+    )
+
+    return fs.existsSync(typePath)
```
seveibar: can you create a class `FilesystemTypesHandler` that is initialized in the DevServer

Classes tend to get massive, let's nip it in the bud and organize the code initially so that we don't have files over 200 lines

seveibar (commented):
looking good, see comment and also i think we need a test

cli/dev/DevServer.ts
ArnavK-09: done

ArnavK-09:
> looking good, see comment and also i think we need a test

wrote, tests, implemented class, anything else left to merge?

thanks

tests/cli/dev/dev.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // Simulate file change with new import
+  const updatedContent = `
+  import { useUsbC } from "@tsci/seveibar.smd-usb-c"
+  export const MyCircuit = () => <></>
+  `
+  fs.writeFileSync(`${tempDirPath}/snippet.tsx`, updatedContent)
+
+  // Trigger file change handler
+  await devServer.handleFileChangedOnFilesystem(`${tempDirPath}/snippet.tsx`)
```
seveibar: this is a bit weird, shouldn't this be automatically called when you update the filesystem? We're not testing the filesystem watcher if we call the dev server manually

seveibar: could you avoid calling this function and instead waiting on the type file in a loop here instead?

ArnavK-09: i have added loop to wait for node module to load as per review
is there anything else left for merge?
thanks

seveibar (approved):

PR #37, opened 2025-01-23
fix: set default port to 3020

krushnarout:
/claim #36
fixes #36 

https://github.com/user-attachments/assets/12ddd2da-ef0d-421a-b3eb-b6f422263009

PR #38, opened 2025-01-23
feat(dev-server): Update default port to 3020 with conflict handling

### LINKED ISSUE #36: Set the default port to 3020. If there is an existing service running on that port, increment the port by 1 (each time)
Currently the port is set to 3000, but 3020 is the standard port for our CLI

Make sure to write a test

/bounty $5

k0msenapati:
/claim #36
Closes #36

### DIFF
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -29,9 +29,9 @@ export class DevServer {
   eventsWatcher?: EventsWatcher
   /**
    * A ky instance that can be used to communicate with the file server and
-   * event bus
+   * event bus. Only available after server starts.
    */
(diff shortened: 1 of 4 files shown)


seveibar (changes requested):
pretty good but let's tweak it to simplify the code.

- Only initialize fsKy once, this means it will be null prior to the server starting (if this is messy, do an any case `null as any` in the constructor)
- Determine if the port is free prior to the `createHttpServer` function, it's very confusing how you pass in a port that may not be used
- Use a library or generally simplify the port selection. It's too complex.
- In the test, don't start two dev server instances, just bind an http server to 3020 and start a single dev server and ping it at 3021 or whatever to make sure it found a port ok

seveibar:
This approach is better https://github.com/tscircuit/cli/pull/37

PR #39, opened 2025-01-23
feat: tsci auth set-token $TOKEN, tsci auth print-token

### LINKED ISSUE #35: `tsci auth set-token $TOKEN`, `tsci auth print-token`
Used when someone wants to get their token for making API calls against the registry API, or explicitly set a token

Make sure to write a test

/bounty $10

ArnavK-09:
image

/claim #35 
trying to resolve #35

### DIFF
--- a/cli/auth/print-token/register.ts
+++ b/cli/auth/print-token/register.ts
@@ -1,0 +1,14 @@
+import type { Command } from "commander"
+import { cliConfig } from "lib/cli-config"
+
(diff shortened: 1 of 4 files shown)


seveibar (approved):
probably ok because it's so simple but you can actually test CLI commands easily!!!

ArnavK-09:
> Quote reply

added some tests, are they alright
can it be merged?

thanks

seveibar (approved):
/tip $5 nice work!

PR #40, opened 2025-01-25
fix: add registry while cloning

### LINKED ISSUE #14: Clone should automatically place an `.npmrc` that sets up the registry
```
@tsci:registry=https://npm.tscircuit.com
```

k0msenapati:
Closes #14


seveibar (approved):
/tip $3

PR #45, opened 2025-01-31
fix: endpoint types issues

k0msenapati:


seveibar (approved):
THANK YOU!!! /tip $10

PR #51, opened 2025-02-01
add indication in the browser and cli that circuit is currently loading

### LINKED ISSUE #29: give indication in the browser and the cli that the circuit is currently loading

Anshgrover23:
/claim #29 
Fixes #29 

### Reference video:

https://github.com/user-attachments/assets/ec3be038-54f5-4ea7-88c5-0714299a4702

### DIFF
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -9,7 +9,6 @@ import fs from "node:fs"
 import type { FileUpdatedEvent } from "lib/file-server/FileServerEvent"
 import * as chokidar from "chokidar"
 import { FilesystemTypesHandler } from "lib/dependency-analysis/FilesystemTypesHandler"
-import { WebSocketServer } from "ws"
 
 export class DevServer {
   port: number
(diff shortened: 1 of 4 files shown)


ShiboSoftwareDev (changes requested):
The issue is not for a loading screen, When the server refreshes and the circuit is still rendering you should indicate to the user that the circuit is rendering, because currently there is no feedback

ShiboSoftwareDev:
@Anshgrover23 
something like this:
image
or this:
image

seveibar (changes requested):
no SSE, you should propose an approach in the issue, the whole reason we have a standard file server is because the server was getting too custom and complicated in the old CLI. The server must remain incredibly simple with just and event bus and api for updating files.

If we're in the process of rendering, you could simply emit an event `CIRCUIT_RENDERING_STARTED` and `CIRCUIT_RENDERING_FAILED` or `CIRCUIT_RENDERING_FINISHED` (then subscribe to these in the browser)

PR #53, opened 2025-02-02
add tsci auth logout

### LINKED ISSUE #52: Implement `tsci auth logout`

Anshgrover23:
Fixes #52 
image

### DIFF
--- a/lib/cli-config/index.ts
+++ b/lib/cli-config/index.ts
@@ -20,7 +20,7 @@ export const setSessionToken = (token: string) => {
   cliConfig.set("githubUsername", decoded.github_username)
 }
 
-export const clearSessionToken = () => {
+export const clearSession = () => {
   cliConfig.delete("sessionToken")
   cliConfig.delete("githubUsername")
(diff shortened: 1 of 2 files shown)


lib/cli-config/index.ts
```diff
@@ -20,6 +20,11 @@ export const setSessionToken = (token: string) => {
   cliConfig.set("githubUsername", decoded.github_username)
 }
 
+export const clearSessionToken = () => {
```
ArnavK-09:
```suggestion
export const clearSession = () => {
```

Anshgrover23: @ArnavK-09 i think it is correct name because  CC: @seveibar as i am not good in naming?
image

seveibar: Yea @ArnavK-09 ’s suggestion is good, your function was misleading because if you had to guess from the name what it would do, it would be unclear that it cleared the github username

seveibar (approved):

Issue #30, opened 2025-01-20
Make "Save to Snippets" button work


seveibar:
/bounty $10

ArnavK-09:
/attempt #30

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/ArnavK-09) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @ArnavK-09 | 20 tscircuit bounties | <div align="center">TypeScript, JavaScript, <br />HTML & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm65cj530000sk1037gjqjj7s/cancel-attempt) |

PR #54, opened 2025-02-06
feat: save to snippet

ArnavK-09:
/claim #30


seveibar (commented):
this is nice but can we avoid calling out to the CLI? There should be a save function right?

You should also be able to test this by emitting a REQUEST_TO_SAVE event, this is a really cool way of testing (event bus testing) that i would definitely try out!!!

increasing bounty on original issue to $18

lib/server/EventsRoutes.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+      }
+    }
+  }
+  "api/events/list": {
+    GET: {
+      responseJson: {
+        event_list: Array<{
+          event_id: string
+          event_type:
```
Anshgrover23, line 20: @ArnavK-09 i think this should be go in seperate files like this file_events lib/file-server/FileServerEvent.ts see reference file.
cc: @seveibar

Anshgrover23 (changes requested):

lib/server/EventsRoutes.ts
ArnavK-09: > @ArnavK-09 i think this should be go in seperate files like this file_events lib/file-server/FileServerEvent.ts see reference file.
> cc: @seveibar 

Confirm it fast I'm going to close my laptop

seveibar: Eventually but this is fine for now

tests/test5-dev-server-save-snippet.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      resolve()
+      cliConfig.set("sessionToken", sessionToken)
+      devServer.fsKy.post("api/events/create", {
+        json: { event_type: "REQUEST_TO_SAVE_SNIPPET" },
+      })
+    })
+  })
+
+  await Bun.sleep(4000)
+  // Wait for the REQUEST_TO_SAVE_SNIPPET event to be detected
```
seveibar, line 82: Sleeps slowly make test suites take forever, i would write a function “waitUntilEvent” with a timeout instead

you can also use promise.race to wait for any of a set of possible events

seveibar (commented):
Pretty good, minor point on test to keep suite fast

tests/test5-dev-server-save-snippet.test.ts
ArnavK-09: Oh I forgot to remove it, test would work even without it, lemme remove it

tests/test5-dev-server-save-snippet.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const sessionToken = cliConfig.get("sessionToken")
+  cliConfig.delete("sessionToken")
+
+  const snippetSavedPromise = new Promise<void>((resolve) => {
+    eventManager.on("SNIPPET_SAVED", () => {
+      resolve()
+    })
+  })
+
+  const snippetSaveFailedPromise = new Promise<void>((resolve) => {
```
seveibar: you can use Promise.withResolvers() as a slightly more readable pattern (this pattern is also fine)

tests/test5-dev-server-save-snippet.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      }),
+    },
+  })
+
+  // Create and start the DevServer instance
+  const devServer = new DevServer({
+    port: devServerPort,
+    componentFilePath: `${tempDirPath}/snippet.tsx`,
+  })
+  await devServer.start()
```
seveibar, line 41: why are you starting a dev server, `getTestServer` already starts a dev server no?

seveibar (commented):
super close but the test is a bit odd because it seems to start two dev servers

this also simplifies your cleanup

also make sure to call `afterAll(() => {})` sooner- if the code fails it might not clean up properly, you generally shouldn't have `afterAll` at the end of a test

tests/test5-dev-server-save-snippet.test.ts
ArnavK-09: that starts snippets server, not devservr that mocks cli dev

ArnavK-09:
> super close but the test is a bit odd because it seems to start two dev servers
> 
> this also simplifies your cleanup
> 
> also make sure to call `afterAll(() => {})` sooner- if the code fails it might not clean up properly, you generally shouldn't have `afterAll` at the end of a test

yup it starts two servers,
fake-snippets to test it code is saved on server or not

and 

dev server that mocks cli dev server


both are neccessary

ArnavK-09:
image

seveibar (approved):
As discussed on discord we will follow up to clean up tests!!

PR #61, opened 2025-02-13
feat: add package.json file generation on init

imrishabh18:


k0msenapati:
@imrishabh18 also do for clone

seveibar (approved):

PR #64, opened 2025-02-13
patch(ux): avoid standlone file read errs

ArnavK-09:


lib/server/createHttpServer.ts
```diff
@@ -14,28 +11,11 @@ export const createHttpServer = async (port = 3020) => {
...
-        return
-      } catch (error) {
-        console.error("Error serving standalone.min.js:", error)
-      }
-
+    if (req.url === "/standalone.min.js") {
       res.writeHead(302, {
-        Location: `https://cdn.jsdelivr.net/npm/@tscircuit/runframe@${pkg.dependencies["@tscircuit/runframe"].replace(/^[^0-9]+/, "")}/dist/standalone.min.js`,
+        Location:
+          "https://cdn.jsdelivr.net/npm/@tscircuit/runframe@0.0.167/dist/standalone.min.js",
```
imrishabh18: Why to hardcode this version `0.0.167`?

imrishabh18 (commented):
I don't think this makes sense to fully depend on the CDN. CDN should be the fallback not the primary option. Regarding the error on `tsci dev` it was resolved cause it was a runframe issue

lib/server/createHttpServer.ts
ArnavK-09: Old commit, updated failed to push

ArnavK-09:
> I don't think this makes sense to fully depend on the CDN. CDN should be the fallback not the primary option. Regarding the error on `tsci dev` it was resolved cause it was a runframe issue

That makes sense, closing pr, just want to confirm if it's published to prod? I'm still getting that error

ArnavK-09:
Runframe isn't found in tsci initt dir

imrishabh18 (approved):
@ArnavK-09 I think this should be good. Try it on your end and resolve the format issue

PR #68, opened 2025-02-13
fix: cli error on `tsci dev`

imrishabh18:
Cli error shown for missing standalone file in runframe repo

### DIFF
--- a/lib/shared/setup-tsci-packages.ts
+++ b/lib/shared/setup-tsci-packages.ts
@@ -5,7 +5,7 @@ import { execSync } from "node:child_process"
 
 export function setupTsciProject(
   directory = process.cwd(),
-  dependencies = ["@types/react", "@tscircuit/core"],
+  dependencies = ["@types/react", "@tscircuit/core", "@tscircuit/runframe"],
 ) {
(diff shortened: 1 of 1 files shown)


lib/shared/setup-tsci-packages.ts
```diff
@@ -5,7 +5,7 @@ import { execSync } from "node:child_process"
 
 export function setupTsciProject(
   directory = process.cwd(),
-  dependencies = ["@types/react", "@tscircuit/core"],
+  dependencies = ["@types/react", "@tscircuit/core", "@tscircuit/runframe"],
```
seveibar, line 8: this shouldn't be a dependency of the user's project. There should be a way for us to build the CLI such that it's not required.

We could have `@tscircuit/cli` as a devDependency which would technically solve this i guess

seveibar: alternatively, we can just prevent the error from printing- i think this is a good solution because it means the user can install a custom runframe version if they want to. In any case let's make sure we don't have anything we don't need as a dependency. Everything here should be a devDependency- i also just realized the variable name might be bad

PR #73, opened 2025-02-14
ci: bump actions versions

ArnavK-09:


.github/workflows/bun-test.yml
```diff
@@ -0,0 +1,24 @@
+name: Bun Test
+
+on:
+  pull_request: ["main"]
```
imrishabh18:
```suggestion
  pull_request:
```

We want it to run for all

imrishabh18 (changes requested):

Issue #90, opened 2025-02-19
If a user runs `tsci init` check that the latest version of the cli is available and offer to install it


seveibar:
/bounty $10

deekshatomer:
/attempt #90

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm7bbb1l40002ii03et4zzjjl/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

AayushSaini101:
/attempt #90

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm7bbb1l40002ii03et4zzjjl/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

MustafaMulla29:
/attempt #90

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/MustafaMulla29) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @MustafaMulla29 | 5 tscircuit bounties | <div align="center">TypeScript, Java, <br />HTML & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm7bbb1l40002ii03et4zzjjl/cancel-attempt) |

MustafaMulla29:
@ArnavK-09 hey, I already submitted a pr. And it's under review

PR #97, opened 2025-03-01
If a user runs tsci init check that the latest version of the cli is available and offer to install it

MustafaMulla29:
/claim #90

### DIFF
--- a/cli/init/register.ts
+++ b/cli/init/register.ts
@@ -1,4 +1,3 @@
-import { SilkscreenCircleProps } from "./../../node_modules/@tscircuit/props/lib/components/silkscreen-circle"
 import type { Command } from "commander"
 import { execSync } from "node:child_process"
 import * as fs from "node:fs"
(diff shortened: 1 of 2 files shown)


techmannih (changes requested):
please take a look, test failed related to init command you need to fix it.
image

MustafaMulla29:
@techmannih test was failing coz checkForUpdates fuction takes user input, so skipped the checkForUpdates function when running tests.

cli/init/register.ts
```diff
@@ -7,6 +7,66 @@ import { generateTsConfig } from "lib/shared/generate-ts-config"
 import { writeFileIfNotExists } from "lib/shared/write-file-if-not-exists"
 import { generateGitIgnoreFile } from "lib/shared/generate-gitignore-file"
 import { generatePackageJson } from "lib/shared/generate-package-json"
+import { detectPackageManager } from "lib/shared/detect-pkg-manager"
+
+const checkForUpdates = async () => {
+  if ((process.env.TSCI_SKIP_UPDATE_CHECK = "true")) {
```
ArnavK-09, line 13:
```suggestion
  if ((process.env.TSCI_SKIP_UPDATE_CHECK == "true")) {
```

ArnavK-09 (changes requested):

Anshgrover23:
@MustafaMulla29 solved via #98

PR #100, opened 2025-03-03
fix: take account of patched version instead of packages version

### LINKED ISSUE #99: `tsci init` always prompts for upgrade, even if you're using the latest version, due to a bug in how the version is determined
CC @ArnavK-09 (I think)

ArnavK-09:
fix #99 

> [!CAUTION]
> export test PROBABLY failing due to core update https://github.com/tscircuit/cli/commit/a797483f03a2c5c98332d1376d82466f205e471c and this pr doesn't changed anything for export , its sperate issue to update snapshots


seveibar (approved):
I think you nailed it

PR #102, opened 2025-03-03
Set up continuous branch release with pkg-pr-new 

### LINKED ISSUE #101: Implement pr cli preview
[pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new)

techmannih:
Set up continuous branch release with pkg-pr-new 

###### resolves #101

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -2,9 +2,6 @@
   "lockfileVersion": 1,
   "workspaces": {
     "": {
-      "dependencies": {
-        "pkg-pr-new": "^0.0.40",
-      },
       "devDependencies": {
         "@babel/standalone": "^7.26.9",
         "@biomejs/biome": "^1.9.4",
@@ -31,6 +28,7 @@
         "ky": "^1.7.4",
         "make-vfs": "^1.0.15",
         "perfect-cli": "^1.0.20",
+        "pkg-pr-new": "^0.0.40",
         "redaxios": "^0.5.1",
         "semver": "^7.6.3",
         "tempy": "^3.1.0",
--- a/bun.lock
+++ b/bun.lock
@@ -2,6 +2,9 @@
   "lockfileVersion": 1,
   "workspaces": {
     "": {
+      "dependencies": {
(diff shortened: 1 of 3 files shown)


techmannih:
I think need to install https://github.com/apps/pkg-pr-new

package.json
```diff
@@ -45,6 +46,7 @@
     "ky": "^1.7.4",
     "make-vfs": "^1.0.15",
     "perfect-cli": "^1.0.20",
+    "pkg-pr-new": "^0.0.40",
```
ArnavK-09: Don't add it as dep, we can use bunx

bun.lock
ArnavK-09, line 1: I don't think we need change in lock, don't install PKG.pr as dep, use bunx

ArnavK-09 (changes requested):
Uhm we need confirmation from other members if we need this

techmannih:
@ArnavK-09 it is same setup in https://github.com/tscircuit/core/pull/424

techmannih:
there using as dep?

techmannih:
you raised issue , I was think need to implement

bun.lock
techmannih: done

ArnavK-09:
> you raised issue , I was think need to implement

Um yeah in my opinion it's needed, as because of it i can spot a mistake in my last pr, still needs confirmation (i said in discord)

@imrishabh18 do we need it?

ArnavK-09:
> @ArnavK-09 it is same setup in https://github.com/tscircuit/core/pull/424

~~Just asking to remove dep, I think it would work without it, since you are using bunx, even in their documentation they didn't asked to add as dep, correct me if I'm wrong
Thanks~~
Ok you did it, thanks

ArnavK-09 (approved):
Although the script in json is surplus, but looks good!

> To customize which package manager is reflected in the comments, use the --packageManager=XYZ flag. XYZ can be one of the following: npm (default), pnpm, yarn, or **bun** ~ from docs

seveibar (approved):

PR #103, opened 2025-03-11
refactor the test setup

imrishabh18:
- refactor the test to a single fixture and remove confusing test fixtures


tests/fixtures/get-cli-test-fixture.ts
```diff
@@ -32,14 +63,28 @@ export async function getCliTestFixture(): Promise<CliTestFixture> {
...
+    // Reset database between tests if needed
+    // This is optional - you might want to keep state between tests
+  })
+  
+  afterAll(cleanup)
 
   return {
     tmpDir,
     runCommand,
+    server,
```
seveibar: Make sure to call this registryServer, all of these should have a registry prefix. The cli also runs a server and we dont want to confuse them

seveibar (approved):

PR #104, opened 2025-03-12
feat private snippet

imrishabh18:


cli/dev/DevServer.ts
```diff
@@ -1,15 +1,15 @@
 import ky from "ky"
-import type { FileServerRoutes } from "lib/file-server/FileServerRoutes"
-import { createHttpServer } from "lib/server/createHttpServer"
-import { EventsWatcher } from "lib/server/EventsWatcher"
+import type { FileServerRoutes } from "../../lib/file-server/FileServerRoutes"
+import { createHttpServer } from "../../lib/server/createHttpServer"
+import { EventsWatcher } from "../../lib/server/EventsWatcher"
```
seveibar: oof why do this? tsconfig should handle this

seveibar (approved):

cli/dev/DevServer.ts
imrishabh18: Ahh vscode was giving swirly lines for it, I will check the config again

Issue #107, opened 2025-03-18
Write test for `tsci clone`


imrishabh18:
Write a test for `tsci clone` to verify if the packages are getting cloned with the correct file structure and files, probably it's not. So modify the the `clone` function to fix it.

/bounty $4

deekshatomer:
/attempt #107


<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm8eigoos0002le03ozkaeeac/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

PR #111, opened 2025-03-20
test(cli): add tests for clone command functionality

ArnavK-09:
/claim #107

### DIFF
--- a/tests/cli/clone/clone.test.ts
+++ b/tests/cli/clone/clone.test.ts
@@ -1,33 +1,29 @@
 import { expect, test } from "bun:test"
 import { join } from "node:path"
 import { getCliTestFixture } from "../../fixtures/get-cli-test-fixture"
-import { readdirSync } from "node:fs"
 
 test("clone command fetches and creates package files correctly", async () => {
   const { tmpDir, runCommand } = await getCliTestFixture()
+  const { stdout } = await runCommand("tsci clone author/snippet")
 
-  const { stdout, stderr } = await runCommand("tsci clone author/snippet")
-expect(stdout).toContain("Cloning author/snippet...")
-
-
-  // Verify directory structure
   const projectDir = join(tmpDir, "author.snippet")
   const indexPath = join(projectDir, "index.tsx")
   const tsconfigPath = join(projectDir, "tsconfig.json")
   const npmrcPath = join(projectDir, ".npmrc")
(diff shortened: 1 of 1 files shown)


tests/cli/clone/clone.test.ts
```diff
@@ -0,0 +1,42 @@
+import { expect, test } from "bun:test"
+import { join } from "node:path"
+import { getCliTestFixture } from "../../fixtures/get-cli-test-fixture"
+
+test("clone command fetches and creates package files correctly", async () => {
```
imrishabh18, line 5: We don't want to make a call to the actual registry for tests, the test should be against the dev server. Something like `init -> push -> clone` to verify.

seveibar: i think this is auto-setup inside the getCliTestFixture?

seveibar (approved):

tests/cli/clone/clone.test.ts
imrishabh18: iirc the db is empty with on packages to verify the clone command

seveibar: oh so basically the server either needs a seed, or he has to push before cloning

seveibar: it's probably not a bad idea to have a seed in the fake server for the sake of making other tests that need a package easier to write imo

ArnavK-09: Aren't these public snippets pre stored to fake snippets db?

https://github.com/tscircuit/tscircuit.com/blob/main/fake-snippets-api%2Flib%2Fdb%2Fautoload-snippets.json

seveibar: I think that's only for the dev server, although it's not a bad idea to bake them/something in. Even something really simple (you could also just manually make some API requests inside the test fixture to do it)

seveibar: one way to verify is just console.log(ky.get(`${apiServerUrl}/packages/list`)) in your test

apiServerUrl is returned from the test fixture fn

ArnavK-09: Already tried that but resulting in unusual error in fake snippets ig something is wrong in api logic
Needs triage

seveibar: kk we will fix in a future PR. definitely bountiable

PR #112, opened 2025-03-20
stylize the tsci dev command

seveibar:
old:

image

new:

image


ArnavK-09:
As dep or dev dep?

seveibar:
> As dep or dev dep?

Good point, forgot we were bundling this project (we should really have a CI rule for this!)

Issue #114, opened 2025-03-20
Exports don't work in browser


seveibar:
![Image](https://github.com/user-attachments/assets/611937e7-2d9e-4884-84c5-e334944f2b80)

/bounty $10

seveibar:
@ArnavK-09 another good one

ArnavK-09:
Duplicate of #94

ArnavK-09:
> [!NOTE]
>
> Depends on https://github.com/tscircuit/runframe/pull/355
> And also need to update runframe version later on

PR #115, opened 2025-03-21
feat: support for export events + unite export sys

ArnavK-09:
PS- Maybe need some patches after merging, works fine as of now
tests failing isnt because of this pr its in prod code, needs discussion and fix in fake-snippets

https://github.com/user-attachments/assets/1878d4c0-c24a-4815-b5f8-90002a787609

image

##### /claim #114


ArnavK-09:
with batch to https://github.com/tscircuit/runframe/pull/355

seveibar (changes requested):
I think you should create the file in a temporary directory or the VFS and have the browser iniaite a download of it. The user should just get a download in their browser

ArnavK-09:
https://github.com/user-attachments/assets/500269ad-9d09-4908-a896-4e33006cdf78

cli/dev/DevServer.ts
```diff
@@ -191,6 +194,43 @@ circuit.add(<MyCircuit />)
...
+      },
+      onSuccess: (response: {
+        fileName: string
+        mimeType: string
+        binaryData: Buffer
+      }) => {
+        postEvent("EXPORT_CREATED", {
+          fileName: response.fileName,
+          mimeType: response.mimeType,
+          binaryData: response.binaryData.toString("base64"),
```
seveibar: we don't want to use `binaryData`, it will crowd the event bus and cause issues. Create a file in the vfs via `/api/files/create` with a path we can ignore when building circuits e.g. `exports/gerbers.zip` then add a `download_url` or `download_path` to the event

seveibar: RunFrame will need to know to ignore all `exports/*` paths

seveibar (approved):

PR #123, opened 2025-03-23
Implement the import tscircuit.com component from the import dialog

### LINKED ISSUE #121: Implement the import tscircuit.com component from the import dialog
This is equivalent to running `tsci add author/component-name`

/bounty $10

Anshgrover23:
/claim #121
Fixes #121

### DIFF
--- a/cli/add/register.ts
+++ b/cli/add/register.ts
@@ -39,16 +39,17 @@ export const registerAdd = (program: Command) => {
 
       // Check if package exists in registry
       const ky = getKy()
-      try {
-        await ky
-          .get("packages/get", {
-            searchParams: { name: componentPath },
-          })
-          .json()
-      } catch (error) {
-        process.stderr.write(`Package ${componentPath} not found in registry\n`)
-        process.exit(1)
-      }
+      await ky
+        .get("packages/get", {
+          searchParams: { name: componentPath },
+        })
+        .json()
+        .catch(() => {
+          process.stderr.write(
+            `Package ${componentPath} not found in registry\n`,
+          )
+          process.exit(1)
+        })
 
       // Read and update package.json
       const packageJsonPath = path.join(workingDir, "package.json")
--- a/cli/add/register.ts
+++ b/cli/add/register.ts
@@ -1,0 +1,103 @@
+import type { Command } from "commander"
+import { detectPackageManager } from "lib/shared/detect-pkg-manager"
+import { execSync } from "node:child_process"
+import * as fs from "node:fs"
+import * as path from "node:path"
+import { getKy } from "lib/registry-api/get-ky"
+
+interface PackageJson {
+  dependencies: Record<string, string>
+}
+
+export const registerAdd = (program: Command) => {
+  program
+    .command("add")
+    .description("Add a component from tscircuit.com")
+    .argument(
+      "<component>",
+      "Component to add (e.g. @tsci/author.component-name)",
+    )
+    .action(async (componentPath: string) => {
+      const workingDir = process.cwd()
+
+      // Ensure the package name is in the correct format
+      if (!componentPath.startsWith("@tsci/")) {
+        process.stderr.write(
+          "Invalid component path. Package name must start with @tsci/\n",
+        )
+        process.exit(1)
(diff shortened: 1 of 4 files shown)


cli/add/register.ts
```diff
@@ -0,0 +1,103 @@
...
+      try {
+        await ky
+          .get("packages/get", {
+            searchParams: { name: componentPath },
+          })
+          .json()
+      } catch (error) {
+        process.stderr.write(`Package ${componentPath} not found in registry\n`)
+        process.exit(1)
+      }
```
ArnavK-09: Should be either .catch() if asynchronous or synchronous if try-catch

ArnavK-09 (changes requested):

ArnavK-09 (changes requested):
Ps - I guess the main issue also includes to integrate import dialog from runframe ig, needs confirmation

cli/add/register.ts
```diff
@@ -0,0 +1,104 @@
...
+      // Read and update package.json
+      const packageJsonPath = path.join(workingDir, "package.json")
+      let packageJson: PackageJson = { dependencies: {} }
+      if (fs.existsSync(packageJsonPath)) {
+        packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"))
+      }
+
+      // Add the package to dependencies
+      packageJson.dependencies[componentPath] = "latest"
+      fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2))
```
seveibar, line 63:
```suggestion
```

seveibar: this is what `bun add` does

cli/add/register.ts
```diff
@@ -0,0 +1,104 @@
...
+        // Install the package using the detected package manager
+        const packageManager = detectPackageManager()
+        const installCommand =
+          packageManager === "yarn"
+            ? `yarn add ${componentPath}`
+            : packageManager === "pnpm"
+              ? `pnpm add ${componentPath}`
+              : packageManager === "bun"
+                ? `bun add ${componentPath}`
+                : `npm install ${componentPath}`
```
seveibar, line 76: this code keeps being repeated and it's a bad pattern, can you create an interop like so:

```
const npm = getInteropPackageManager()

npm.install.execSync(componentPath, {
  // ...
})
```

tests/cli/add/add.test.ts
```diff
@@ -14,6 +15,34 @@ test.skip("add command installs package and updates .npmrc", async () => {
...
+    is_footprint: false,
+    is_source_from_github: false,
+    latest_package_release_id: "test-release-id",
+    latest_version: "1.0.0",
+    ai_description: null,
+    snippet_type: "board",
+  })
+
   // Run add command
   const { stdout } = await runCommand("tsci add @tsci/example-package")
```
seveibar, line 47: it seems like this already works b/c test? Why is there already a test for this? Was the test broken

cli/add/register.ts
```diff
@@ -0,0 +1,104 @@
...
+  dependencies: Record<string, string>
+}
+
+export const registerAdd = (program: Command) => {
+  program
+    .command("add")
+    .description("Add a component from tscircuit.com")
+    .argument(
+      "<component>",
+      "Component to add (e.g. @tsci/author.component-name)",
```
seveibar, line 18:
```suggestion
      "Component to add (e.g. username/component-name)",
```

cli/add/register.ts
```diff
@@ -0,0 +1,104 @@
...
+    )
+    .action(async (componentPath: string) => {
+      const workingDir = process.cwd()
+
+      // Ensure the package name is in the correct format
+      if (!componentPath.startsWith("@tsci/")) {
+        process.stderr.write(
+          "Invalid component path. Package name must start with @tsci/\n",
```
seveibar, line 26: we should adjust the package name automatically if it doesn't have the prefix. We accept with or without prefix

If the user doesn't specify an author, we should prompt them to select one of the available packages (if there are multiple) listed by star count (+$8)

PR #135, opened 2025-03-26
refactor: remove unused export functionality

ArnavK-09:
should i also remove export function and put logic directory into cli command file like before?

image


seveibar (changes requested):
😭

seveibar (approved):
sorry 😭

Issue #136, opened 2025-03-27
Make title of page the circuit name in `tsci dev`


seveibar:
![Image](https://github.com/user-attachments/assets/2380b753-1fd5-46b5-ab67-192570baff2d)

/bounty $8

gurjeetsinghvirdee:
/attempt #136

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm8qz6i7u0004kv03iufobsy9/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

MustafaMulla29:
/attempt #136

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/MustafaMulla29) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @MustafaMulla29 | 11 tscircuit bounties | <div align="center">TypeScript, Java, <br />JavaScript & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm8qz6i7u0004kv03iufobsy9/cancel-attempt) |

ArnavK-09:
solved via https://github.com/tscircuit/runframe/pull/473

PR #138, opened 2025-03-27
Make title of page the circuit name in tsci dev

MustafaMulla29:
/claim #136

### DIFF
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -60,7 +60,11 @@ export class DevServer {
   }
 
   async start() {
-    const { server } = await createHttpServer(this.port)
+    const circuitName = path.basename(
+      this.componentFilePath,
+      path.extname(this.componentFilePath),
+    )
+    const { server } = await createHttpServer(this.port, circuitName)
     this.httpServer = server
 
     this.eventsWatcher = new EventsWatcher(`http://localhost:${this.port}`)
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -19,7 +19,6 @@ export class DevServer {
    * The path to a component that exports a <board /> or <group /> component
    */
   componentFilePath: string
-
   projectDir: string
 
   /**
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -60,10 +60,22 @@ export class DevServer {
   }
 
   async start() {
-    const circuitName = path.basename(
-      this.componentFilePath,
-      path.extname(this.componentFilePath),
-    )
+    const componentDir = path.dirname(this.componentFilePath)
+    const packageJsonPath = path.resolve(componentDir, "package.json")
+
+    let circuitName = "TSCircuit"
+
+    if (fs.existsSync(packageJsonPath)) {
+      try {
+        const packageJson = JSON.parse(
+          fs.readFileSync(packageJsonPath, "utf-8"),
+        )
+        circuitName = packageJson.name || circuitName
+      } catch (error) {
+        console.error("Failed to read package.json:", error)
+      }
+    }
+
     const { server } = await createHttpServer(this.port, circuitName)
     this.httpServer = server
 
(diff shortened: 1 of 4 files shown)


MustafaMulla29:
image

cli/dev/DevServer.ts
```diff
@@ -60,7 +60,11 @@ export class DevServer {
   }
 
   async start() {
-    const { server } = await createHttpServer(this.port)
+    const circuitName = path.basename(
+      this.componentFilePath,
+      path.extname(this.componentFilePath),
+    )
```
imrishabh18: This is not the circuit name, it will always be `index` as the default file is `index.tsx`. Get the circuit name from the `package.json`.

lib/site/getIndex.ts
```diff
@@ -1,12 +1,16 @@
...
   return `<html>
     <head>
+      <title>${circuitName}</title>
     </head>
     <body>
       <script>
         window.TSCIRCUIT_USE_RUNFRAME_FOR_CLI = true;
+        if (!window.location.pathname.includes("${circuitName}")) {
+          window.history.replaceState({}, "", "/${circuitName}");
+        }
```
imrishabh18: I don't know what are you trying to achieve from this.

imrishabh18 (changes requested):

lib/site/getIndex.ts
MustafaMulla29: @imrishabh18 this will append the url with /circuitName. Isn't it?. I'll remove it if you want

imrishabh18: I can figure that out, but why do we need to append it to the url?

MustafaMulla29: There is two arrows in the issue image, so I thought I have to append it😅. Ok my bad I'll remove it

MustafaMulla29:
image


@imrishabh18 is this ok?

lib/server/createHttpServer.ts
```diff
@@ -50,8 +50,8 @@ export const createHttpServer = async (port = 3020) => {
       return
     }
 
-    if (url.pathname === "/") {
-      const html = await getIndex()
+    if (url.pathname === "/" || url.pathname === `/${circuitName}`) {
+      const html = await getIndex(circuitName)
```
imrishabh18: Remove this change as well

imrishabh18 (approved):
See comments

lib/server/createHttpServer.ts
MustafaMulla29: @imrishabh18 sure

cli/dev/DevServer.ts
```diff
@@ -60,7 +59,23 @@ export class DevServer {
   }
 
   async start() {
-    const { server } = await createHttpServer(this.port)
+    const componentDir = path.dirname(this.componentFilePath)
+    const packageJsonPath = path.resolve(componentDir, "package.json")
+
+    let circuitName = "TSCircuit"
```
seveibar:
```suggestion
    let circuitName = "tsci dev"
```

cli/dev/DevServer.ts
```diff
@@ -60,7 +59,23 @@ export class DevServer {
...
+    if (fs.existsSync(packageJsonPath)) {
+      try {
+        const packageJson = JSON.parse(
+          fs.readFileSync(packageJsonPath, "utf-8"),
+        )
+        circuitName = packageJson.name || circuitName
+      } catch (error) {
+        console.error("Failed to read package.json:", error)
+      }
+    }
```
seveibar: isolate all this circuit name logic into another function, we need to keep the DevServer.ts file very readable

seveibar (changes requested):
it can't be static, it should be set dynamically in RunFrame

we are going to have support for changing circuits within a project later. The title should be the exported component name, or as a fallback the package json name

lib/server/createHttpServer.ts
```diff
@@ -51,7 +51,7 @@ export const createHttpServer = async (port = 3020) => {
     }
 
     if (url.pathname === "/") {
-      const html = await getIndex()
+      const html = await getIndex(circuitName)
```
seveibar: this is a bad positional parameter, if the parameter isn't obvious make it a named object
```suggestion
      const html = await getIndex({ circuitName })
```

MustafaMulla29:
image

@seveibar is this correct?

seveibar (changes requested):
no sorry, this issue requires a change in RunFrame (RunFrame should manage the title of the page by looking for the best circuit name)

MustafaMulla29:
@seveibar so there is no change in this repo?

PR #140, opened 2025-03-28
fix: show on tscircuit.com on `tsci push` (WIP)

imrishabh18:


seveibar:
This has been rolled into #146

PR #164, opened 2025-04-16
feat(cli): add directory confirmation prompt for init command

ArnavK-09:
tests failing proabbly due to  #162, but is not related to this pr

Add a confirmation prompt to the `tsci init` command when no directory is provided. This ensures the user explicitly confirms initializing a project in the current directory. Additionally, update the test to use a specific project directory for better clarity and consistency.

### DIFF
--- a/cli/init/register.ts
+++ b/cli/init/register.ts
@@ -1,5 +1,4 @@
 import type { Command } from "commander"
-import { execSync } from "node:child_process"
 import * as fs from "node:fs"
 import * as path from "node:path"
 import { setupTsciProject } from "lib/shared/setup-tsci-packages"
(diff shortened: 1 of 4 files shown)


cli/init/register.ts
```diff
@@ -22,6 +24,15 @@ export const registerInit = (program: Command) => {
     .action(async (directory?: string) => {
       await checkForTsciUpdates()
 
+      if (!directory) {
+        const continueInCurrentDirectory = await askConfirmation(
+          "Do you want to initialize a new project in the current directory?",
+        )
+        if (!continueInCurrentDirectory) {
+          process.exit(0)
```
seveibar: I would avoid using process.exit, just throw an error. Anything with process.exit messes up tests

seveibar (approved):

PR #176, opened 2025-04-20
feat(cli): add upgrade command to check for CLI updates

ArnavK-09:
This commit introduces a new `upgrade` command that allows users to check for the latest version of the CLI. If the user is already on the latest version, a message is displayed indicating no update is needed. This enhances the CLI by providing a built-in mechanism for version management.


seveibar (approved):

seveibar:
sidenote: askConfirmation should be refactored to use `lib/prompts` because it has better defaults etc.

Issue #182, opened 2025-04-23
On `tsci clone <author-name>/<package-name>` the cloned directory name should not have `author-name`


imrishabh18:
![Image](https://github.com/user-attachments/assets/5ab11cd4-1388-44f8-84fd-a5ce11d50c04)

The package name should be just `open-closed`

PR #183, opened 2025-04-23
feat: `tsci config set` && refactor(clone): improve error message, author name customization 

ArnavK-09:
The directory path for cloned snippets no longer includes the author prefix, simplifying the structure and making it more consistent with user expectations. This change also updates the corresponding test cases to reflect the new directory naming convention.

Introduce a new configuration option `alwaysCloneWithAuthorName` to control whether the author name is included in the cloned snippet directory by default. This allows users to set a global preference for including the author name, reducing the need to specify the `--include-author` flag repeatedly.

Additionally, improve error handling for the clone command by providing a more user-friendly message when a snippet is not found. Register a new `config set` command to allow users to modify configuration values programmatically.
###### /claim #182

### DIFF
--- a/cli/clone/register.ts
+++ b/cli/clone/register.ts
@@ -14,96 +14,106 @@ export const registerClone = (program: Command) => {
       "<snippet>",
       "Snippet to clone (e.g. author/snippetName or https://tscircuit.com/author/snippetName)",
     )
-    .action(async (snippetPath: string) => {
-      // First try to match URL format (strict tscircuit.com only)
-      const urlMatch = snippetPath.match(
-        /^https:\/\/tscircuit\.com\/([^\/]+)\/([^\/]+)\/?$/i,
-      )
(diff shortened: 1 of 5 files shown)


seveibar:
@imrishabh18 so I actually like this feature where we keep the author name in the package, github doesn't do it and that's our reference so maybe it's ok to disable it by default, but for my selfish self we should make it a config setting that i can turn on 😁 

the reason i like it is because the directory structure you create resembles the registry, i'm sure it would be really annoying for teams where everything in the directory has `company.*` though!!!!

seveibar (changes requested):
the flag is OK but can you make it a user setting as well?

also note, `-ia` is not a best-practice flag since in many CLIs it means `-i` and `-a`, i would maybe use `--include-author` / `-a`

seveibar (approved):
Niceee

PR #184, opened 2025-04-30
handle INSTALL_PACKAGE and emit corresponding events

### LINKED ISSUE #134: Handle `INSTALL_COMPONENT` event and emit `COMPONENT_INSTALLED` and `COMPONENT_INSTALL_FAILED`
CC @ArnavK-09 

/bounty $15

ShiboSoftwareDev:
/claim #134
/closes #134


cli/dev/DevServer.ts
```diff
@@ -169,9 +175,37 @@ export class DevServer {
...
+        throwHttpErrors: false,
+      })
+    }
+
+    try {
+      console.log(`Installing package ${full_package_name}…`)
+      const cmd = `tsci add ${full_package_name}`
+
+      // Synchronously run the install
+      execSync(cmd, { stdio: "pipe" })
```
seveibar: Can you call the installation method directly? Going out to shell breaks across operating systems a lot so we want to do it as little as possible.

There should be a function already inplemented for it but if not you may need to break out the add method into a separate func

seveibar (commented):
Recommend using shell calls as little as possible!

ShiboSoftwareDev:
made the logic into a reusable function

seveibar (approved):
very nice

Issue #174, opened 2025-04-20
Either entrypoint or mainComponentPath must be provided (no index file, could not infer entrypoint)


ShiboSoftwareDev:
this is the first thing I get when cloning the cli and running bun start

![Image](https://github.com/user-attachments/assets/2181dc01-dee2-4b3e-9a40-369348eff3b0)

PR #186, opened 2025-05-02
feat(DevServer): add event creation for component file

ArnavK-09:
This commit introduces a new event creation when a component file is detected during the `upsertInitialFiles` process. The event type `SET_COMPONENT_FILE` is sent to the server with the component file path, enhancing the ability to track and manage component files in the development environment.

related #174


seveibar (commented):
Confused why we need this, whenever a file is upserted we automatically emit an event already. If i have a bit more context maybe ill understand

PR #189, opened 2025-05-07
feat(cli): enhance package search with interactive selection and installation

ArnavK-09:
https://github.com/user-attachments/assets/661cfecc-0803-478d-9ddd-365e599bcc80

Added prompts for users to select a package from search results and confirm installation. This improves user experience by allowing interactive package management directly from the CLI.

### DIFF
--- a/cli/search/register.ts
+++ b/cli/search/register.ts
@@ -47,7 +47,7 @@ export const registerSearch = (program: Command) => {
       const { selectedPackage } = await prompts({
         type: "select",
         name: "selectedPackage",
-        message: "Select a package to install:",
+        message: "Select a package to add:",
         choices,
         initial: 0,
       })
--- a/cli/search/register.ts
+++ b/cli/search/register.ts
@@ -1,6 +1,8 @@
 import type { Command } from "commander"
 import { getRegistryApiKy } from "lib/registry-api/get-ky"
 import kleur from "kleur"
+import prompts from "prompts"
+import { addPackage } from "lib/shared/add-package"
 
 export const registerSearch = (program: Command) => {
   program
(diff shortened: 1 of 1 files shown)


seveibar (approved):
Much much better cc @Anshgrover23 

Also if the user doesnt type in a query they should be able to interactively search (future pr?)

divanshu-go:
@seveibar I don't think search command should prompt users for literally anything.
It depends on the user what they want to do with found snippets
They can either clone it using `tsci clone` or import it into their existing snippets using `tsci add`

So making search command interactive just takes away it minimalism and make it harder to use.

Anshgrover23:
@divanshu-go agree with you.

divanshu-go:
@ArnavK-09 you should instead add this interactivity to `tsci clone` and `tsci add`
If user don't provide the argument to them
Then we should search packages and prompt them if they want to install that one or either they could select multiple packages

ArnavK-09:
> @ArnavK-09 you should instead add this interactivity to `tsci clone` and `tsci add` If user don't provide the argument to them Then we should search packages and prompt them if they want to install that one or either they could select multiple packages

diy?

divanshu-go:
> > @ArnavK-09 you should instead add this interactivity to `tsci clone` and `tsci add` If user don't provide the argument to them Then we should search packages and prompt them if they want to install that one or either they could select multiple packages
> 
> diy?

did you meant, `do it yourself` ?

ArnavK-09:
> > > @ArnavK-09 you should instead add this interactivity to `tsci clone` and `tsci add` If user don't provide the argument to them Then we should search packages and prompt them if they want to install that one or either they could select multiple packages
> > 
> > 
> > diy?
> 
> did you meant, `do it yourself` ?

im unable to understand what you mean, this looks minimal, no 
?

seveibar:
@divanshu-go generally agree that `tsci add` and `tsci clone` could/should be interactive, but I don't want to add complexity to the CLI when there's so much other stuff that's broken at the moment- it's ok that we're isolating to search.

I generally agree with your point but having search be interactive is an OK short term compromise. We should focus on usability/capability over building a TUI rn.

PR #190, opened 2025-05-07
add interactivity of searching to both add and clone commands

Anshgrover23:
image


seveibar:
We aren't using inquirer, we're using prompts

I think clone and add should match git/npm, it's not as important for them to be interactive, it's a good feature idea but not something we need rn

The way that you search is totally wrong. It's not a good TUI for searching. As a user types, options should appear. To do it properly would probably require ink

PR #194, opened 2025-05-11
fix(cli): prevent --version/-V from printing version with subcommands

Anshgrover23:
**Before:**

```powershell
PS C:\Users\ansh\cli> tsci
√ Choose command » tsci login
> tsci login --version false --V false
0.1.100
PS C:\Users\ansh\cli>
```

**After:**

```powershell
PS C:\Users\ansh\cli> tsci
√ Choose command » tsci login
> tsci login
Already logged in! Use 'tsci logout' if you need to switch accounts.
PS C:\Users\ansh\cli>
```

### DIFF
--- a/cli/main.ts
+++ b/cli/main.ts
@@ -47,7 +47,11 @@ registerUpgradeCommand(program)
 registerSearch(program)
 
 // Manually handle --version, -v, and -V flags
-if (process.argv.includes('--version') || process.argv.includes('-v') || process.argv.includes('-V')) {
+if (
+  process.argv.includes("--version") ||
+  process.argv.includes("-v") ||
+  process.argv.includes("-V")
+) {
   console.log(getVersion())
   process.exit(0)
 }
--- a/cli/main.ts
+++ b/cli/main.ts
@@ -21,10 +21,7 @@ import { registerSearch } from "./search/register"
 
 export const program = new Command()
 
-program
-  .name("tsci")
-  .description("CLI for developing tscircuit snippets")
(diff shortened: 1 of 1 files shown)


Anshgrover23:
@seveibar 
earlier when the user only run `tsci`
the interactive promt started 
user choose command
the command run but there were two problems

1. see that the output is just the current version of cli
```bash
PS C:\Users\ansh\cli> tsci
√ Choose command » tsci login
> tsci login --version false --V false
0.1.100
PS C:\Users\ansh\cli>
```
2. see there were reduntant flags ```--version false --V false```

imrishabh18 (approved):

ArnavK-09 (approved):
Did you try if this thing working fine?

https://github.com/tscircuit/cli/blob/97e90dce3c21d887aecf33090770d04612409037/lib/shared/check-for-cli-update.ts#L11-L12

Or you need to update it using getVersion

Rest lgtm

seveibar (changes requested):
this breaks `tsci --version`

Anshgrover23:
now `tsci --version` works
@seveibar the upgrade commands work as expected
```powershell
PS C:\Users\anshg> tsci upgrade
You are already using the latest version of tsci (v0.1.101)
PS C:\Users\anshg>
```

seveibar (approved):

seveibar:
FWIW breaks any command that uses `-v` as a parameter (which is commonly used to mean `--verbose`)

PR #205, opened 2025-05-23
update packages and the test server (latest schematic-symbols support)

imrishabh18:


seveibar (changes requested):
Super confused, this uses a webworker when it should be using a runner

lib/shared/generate-circuit-json.ts
```diff
@@ -37,7 +38,11 @@ export async function generateCircuitJson({
 }: GenerateCircuitJsonOptions) {
   debug(`Generating circuit JSON for ${filePath}`)
 
-  const runner = new CircuitRunner()
+  const worker = await createCircuitWebWorker({
```
seveibar: This is a major breaking change, nodejs does not have webworker support and webworkers should not be used outside the browser

imrishabh18: Ohhh yeah, but when I am using the latest eval with the runner it was breaking

imrishabh18: Let me look more into this

imrishabh18: @seveibar Getting this trace error when using the latest version of eval

image

PR #209, opened 2025-05-31
Refactor pushSnippet function to remove snippetType prompt and default to "package" for snippet_type

### LINKED ISSUE #208: Remove the option of `Board` from the Package Type
Don't need the prompt of `Package Type` to select on `tsci push`. It should always be `Package`

Image

ArnavK-09:
…

fix #208

### DIFF
--- a/lib/shared/push-snippet.ts
+++ b/lib/shared/push-snippet.ts
@@ -164,7 +164,7 @@ export const pushSnippet = async ({
     })
 
   if (!doesPackageExist) {
-    const { createPackage, visibility, snippetType } = await prompts([
+    const { createPackage, visibility } = await prompts([
       {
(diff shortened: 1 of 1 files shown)


seveibar (changes requested):
we should still prompt for this (or is it determined by the server?) boards can't really be imported so should be specially marked no?

seveibar:
I could be wrong btw- should have done comment instead of request change

ArnavK-09:
cc @imrishabh18 ?

imrishabh18:
@ArnavK-09 I don't remember you will have to look deeper!

ArnavK-09:
> @ArnavK-09 I don't remember you will have to look deeper!

🥹

PR #219, opened 2025-06-06
Implement tsci pull to sync cli with registry

imrishabh18:
## Summary
- emit raw error objects from `pullPackage`
- handle plain error objects in `tsci pull` command

## Testing
- `bun run format:check`
- `bun test tests/cli/pull/pull.test.ts` (1 skipped)

------
https://chatgpt.com/codex/tasks/task_b_68435a14eec88327b96aa8c9a6f4fc70

### DIFF
--- a/cli/main.ts
+++ b/cli/main.ts
@@ -15,6 +15,7 @@ import { registerAuthPrintToken } from "./auth/print-token/register"
 import { registerAuthSetToken } from "./auth/set-token/register"
 import { registerPush } from "./push/register"
 import { registerAdd } from "./add/register"
+import { registerPull } from "./pull/register"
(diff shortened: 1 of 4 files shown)


seveibar (changes requested):
I don't want to support this yet. We're not a version control system, we're more like NPM

seveibar:
we would need to error if the user had any local changes etc. npm doesn't provide a way to pull, only a way to publish and install.

seveibar:
we should be more like vercel with `vc` to deploy/publish or npm.

imrishabh18:
Okay, I made changes in the tscircuit.com and wanted that changes in cli. So, I thought that a `pull` will be a good feature. But yeah makes sense to not have a version control system

PR #237, opened 2025-06-19
snapshot command now works in windows machines

ShiboSoftwareDev:
the workflows use linux so this shouldn't introduce regressions if the workflows pass am I right?


seveibar (changes requested):
process.chdir is definitely a hack that we shouldn't need to do

seveibar:
historically fixing windows has often meant importing path/posix

ShiboSoftwareDev:
True, how did i forget about that

seveibar (approved):

PR #241, opened 2025-06-20
refactored config creation logic

ShiboSoftwareDev:


seveibar (changes requested):
a method with effects MUST be named such that the effects are obvious, `getEntrypoint` is not allowed to have effects

seveibar:
I would also say that your title is misleading, you're adding the ability to update a config but the title implies the opposite

seveibar:
WAIT wtf it is already saving a config- totally bork. Let's remove that completely

seveibar:
@ShiboSoftwareDev may require refactoring where we do expect to create a config, but it should never be created from this method, feel free to `git blame` and name and shame the author for that one (maybe it's me but i hope not)....

ShiboSoftwareDev:
you asked for it
image

seveibar (approved):

PR #245, opened 2025-06-21
fixed regression that caused snapshot to check all index file

ShiboSoftwareDev:
this is a regression that caused the command to break because it tried to check irrelevant index files
+ updated tscircuit dep


lib/shared/snapshot-project.ts
```diff
@@ -56,23 +56,13 @@ export const snapshotProject = async ({
...
-      const resolved = path.resolve(projectDir, entry)
-      if (!files.includes(resolved)) {
-        files.unshift(resolved)
-      }
-    }
   }
 
   if (files.length === 0) {
-    return onExit(1)
+    onSuccess(
```
imrishabh18: Why is it `onSucess` but the message doesn't seem like a success condition?

ShiboSoftwareDev: should I just do a console log?

imrishabh18: I think so 🤷🏻‍♂️

imrishabh18: Just double check if there is some method which being called gets the message displayed on the terminal, in place of the console log. Better consistency in the codebase

seveibar (approved):

PR #247, opened 2025-06-22
tscircuit is now a peer dependency

ShiboSoftwareDev:
does it have to also be a dev dep?


ArnavK-09 (approved):

seveibar (commented):
@ShiboSoftwareDev yes also has to be a dev dep, otherwise it will default use the global "tscircuit" installation

PR #249, opened 2025-06-24
props is now a peer dependency

ShiboSoftwareDev:


seveibar (approved):

seveibar (commented):
difficult to implement but technically props should not really be _any dependency_, it is provided by the "tscircuit" package

ShiboSoftwareDev:
yeah, it's in a bunch of deps:
image

PR #250, opened 2025-06-25
build command now ignores circuitJson errors with the --ignore-errors flag

ShiboSoftwareDev:


lib/shared/setup-github-actions.ts
```diff
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
       - run: bun install
-      - run: bunx tsci build
+      - run: bunx tsci build --ignore-errors
```
seveibar: idk i think it's a good idea to generally force circuits to always build without errors

```suggestion
      - run: bunx tsci build
```

seveibar (approved):
you can hint the the user `# --ignore-errors` if you'd like

PR #261, opened 2025-07-08
Add Node snapshot smoke test workflow

seveibar:
## Summary
- add a workflow running a Node-based smoke test for snapshotting
- create a small Node script that snapshots a simple board using the installed `tsci`

## Testing
- `bun run build`
- `npm install -g .`
- `node scripts/node-snapshot-test.js`
- `bun run format`

------
https://chatgpt.com/codex/tasks/task_b_686545bff3a0832e8284c5e69117a2df

### DIFF
--- a/.github/workflows/node-snapshot.yml
+++ b/.github/workflows/node-snapshot.yml
@@ -1,0 +1,28 @@
+name: Node Snapshot Test
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  node-snapshot:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/node-snapshot.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Setup bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+      - name: Install dependencies
+        run: bun install
+      - name: Build and install tsci
+        run: |
+          bun run build
+          npm install -g .
```
ArnavK-09, line 26: Use bun?

seveibar: the point of this test is to make sure that someone who doesn't have bun installed can still run `tsci`, because in the past we had breakage for people who just did `npm install -g tscircuit` and `tsci dev` without ever using bun

PR #279, opened 2025-07-14
replaced deps with the dep "tscircuit"

ShiboSoftwareDev:
I wanted to update circuit-to-svg but I think this is a better solution long term


package.json
```diff
@@ -24,10 +21,8 @@
     "@types/semver": "^7.5.8",
     "bun-match-svg": "^0.0.12",
     "chokidar": "4.0.1",
-    "tsx": "^4.7.1",
     "circuit-json-to-readable-netlist": "^0.0.9",
     "circuit-json-to-simple-3d": "^0.0.3",
-    "circuit-to-svg": "^0.0.164",
```
imrishabh18, line 30: `circuit-to-svg` is not part of `tscircuit` dependency. I think you need to add this back

ShiboSoftwareDev: @imrishabh18 it is, otherwise the tests wouldn't have passed:
image

imrishabh18: Okay, yeah seems like it was added recently. 

@seveibar What was the reason for adding `circuit-to-svg` as a dependency to `tscircuit`? It's just a library which needs to be independent, right?

seveibar: It should be there for the cli to export svgs iirc, tscircuit installs everything needed for the cli

seveibar (approved):

PR #280, opened 2025-07-14
Handle snapshot metadata differences

Abse2001:
## Summary
- ensure snapshot comparison ignores `data-software-used-string` metadata
- add regression test covering metadata differences

## Testing
- `bun test tests/looks-same-metadata.test.ts`
- `bun test` *(fails: basic init, token handling, clone and other CLI tests)*

------
https://chatgpt.com/codex/tasks/task_e_68750bd5d9a883308e39d3277bdb804b

------
the only thing changing is the metadata of core version 

image


seveibar (changes requested):
metadata should not effect looks-same... something else is wrong here i think (this is super sensitive to any id changes)

Abse2001:
@seveibar I will check

PR #281, opened 2025-07-16
Introduce image-based visual diffing for more reliable snapshot verification

Abse2001:


seveibar (approved):

Abse2001:
@seveibar IIUC looks same reads the svg file not svg string so i had to save the svg string to temp file and pass it to looks same to diff

seveibar (commented):
Looks same should accept buffers, you should be able to just use buffers without writing to files, i will check some docs

seveibar (approved):

PR #285, opened 2025-07-20
Better entrypoint resolution with explicit check for provided file name

### LINKED ISSUE #283: dev command should search the directory for circuits and prompt you to choose one

ArnavK-09:


package.json
```diff
@@ -8,6 +8,7 @@
     "@tscircuit/fake-snippets": "^0.0.87",
     "@tscircuit/file-server": "^0.0.24",
     "@tscircuit/math-utils": "^0.0.18",
+    "@tscircuit/props": "^0.0.262",
```
seveibar: this should be provided by the `tscircuit` module

seveibar (changes requested):
i think we should start the server and have the user use runframe to select the entrypoint

seveibar:
Runframe should just open a default file and the user can select a different file if the default file is not what they wanted

seveibar (approved):

seveibar:
this PR has like more commits than lines changed lol

PR #286, opened 2025-07-21
Fix Either entrypoint or mainComponentPath must be provided

### LINKED ISSUE #174: Either entrypoint or mainComponentPath must be provided (no index file, could not infer entrypoint)

ArnavK-09:
fix https://github.com/tscircuit/cli/issues/174

### DIFF
--- a/lib/server/createHttpServer.ts
+++ b/lib/server/createHttpServer.ts
@@ -10,9 +10,11 @@ import runFrameStandaloneBundleContent from "@tscircuit/runframe/standalone" wit
 // @ts-ignore
 import winterspecBundle from "@tscircuit/file-server/dist/bundle.js"
 import { getIndex } from "../site/getIndex"
-import kleur from "kleur"
 
-export const createHttpServer = async (port = 3020) => {
+export const createHttpServer = async (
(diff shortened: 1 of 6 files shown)


lib/server/createHttpServer.ts
```diff
@@ -10,9 +10,11 @@ import runFrameStandaloneBundleContent from "@tscircuit/runframe/standalone" wit
 // @ts-ignore
 import winterspecBundle from "@tscircuit/file-server/dist/bundle.js"
 import { getIndex } from "../site/getIndex"
-import kleur from "kleur"
 
-export const createHttpServer = async (port = 3020) => {
+export const createHttpServer = async (
+  port = 3020,
+  mainComponentPath?: string,
```
seveibar: good time to convert to named arguments, `port` was obvious but the path is not obvious at all

seveibar (commented):
should this be like `defaultMainComponentPath`? Can't the user change it from RunFrame (select other boards/files)?

seveibar (approved):

PR #300, opened 2025-07-28
Update README CLI command

imrishabh18:
## Summary
- clarify that installing `@tscircuit/cli` exposes the `tscircuit-cli` command
- mention that `tsci` comes from the `tscircuit` package
- update examples and help output to use `tscircuit-cli`
- **remove redundant line** telling users to use `tscircuit-cli` if they only install the CLI

## Testing
- `bun test --timeout 20000` *(fails: 1 failed, 13 skipped)*

------
https://chatgpt.com/codex/tasks/task_b_6886ee6472ec8327b3feff33116cbda4

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,20 +6,27 @@ https://github.com/user-attachments/assets/0228e09d-48fc-4bf5-814b-762b60fc35c7
 
 ## Usage
 
+The CLI installed from `@tscircuit/cli` exposes the `tscircuit-cli` command. The
+shorter `tsci` alias is provided by the `tscircuit` package.
+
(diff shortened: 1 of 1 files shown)


seveibar (commented):
I think tscircuit-cli should just be noted as a development detail and it shouldnt be in the readme

imrishabh18:
> I think tscircuit-cli should just be noted as a development detail and it shouldnt be in the readme

Okay, how about adding a command in the README to also install the `tscircuit` package along with the `@tscircuit/cli`. That would make things simpler?

PR #310, opened 2025-08-09
change tscircuit to optional peer dep

ShiboSoftwareDev:


bun.lock
```diff
@@ -48,14 +48,17 @@
         "redaxios": "^0.5.1",
         "semver": "^7.6.3",
         "tempy": "^3.1.0",
-        "tscircuit": "^0.0.574-libonly",
+        "tscircuit": "^0.0.585-libonly",
         "tsx": "^4.7.1",
         "typed-ky": "^0.0.4",
-        "zod": "3",
```
seveibar: !!! zod 4 will break everything

PR #311, opened 2025-08-11
remove tscircuit from peer dep

ShiboSoftwareDev:


bun.lock
```diff
@@ -48,14 +48,11 @@
         "redaxios": "^0.5.1",
         "semver": "^7.6.3",
         "tempy": "^3.1.0",
-        "tscircuit": "^0.0.574-libonly",
```
seveibar: we want this to be libonly to avoid having a circular dependency with @tscircuit/cli (npm ls @tscircuit/cli should find no copies in this repo)

package.json
```diff
@@ -52,14 +52,6 @@
     "looks-same": "^9.0.1",
     "sharp": "^0.34.3"
   },
-  "peerDependencies": {
-    "tscircuit": "*"
-  },
-  "peerDependenciesMeta": {
-    "tscircuit": {
-      "optional": false
```
seveibar: can you set `optional: true` instead? I'm pretty sure your change will cause tscircuit to be bundled

PR #312, opened 2025-08-14
fix: warning message shown on every command you run

imrishabh18:
image


seveibar (approved):
Does this break tsci snapshot? Cc @Abse2001

seveibar:
@imrishabh18 the test failure is legit, this wont work

imrishabh18:
@seveibar Merging this, tests are passing now

PR #319, opened 2025-09-06
support local .obj file loading

### LINKED ISSUE #316: Local `obj` file loading
CC @ShiboSoftwareDev @ArnavK-09 

```tsx
import myObjUrl from "./myObj.obj"

export const MyComponent = () => {
  return (
    <chip
      cadModel={{ objUrl: myObjUrl }}
    />
  )
}
```

How it works:

- Every obj import becomes a URL to a static asset, e.g. https://localhost:3020/....
- Vite does this for images, but i'm not 100% sure how

/bounty $150

ShiboSoftwareDev:
this was possible with the updated made to 3d-viewer and eval
the resistor is a normal .obj file and the other is an OpenCASCADE obj file 
image
/claim https://github.com/tscircuit/cli/issues/316
/closes https://github.com/tscircuit/cli/issues/316


seveibar (approved):
Fwiw we need to create a .d.ts file for declaring obj files are valid to import automatically for the user, this can be done by including the declaration inside the cli and using the tsconfig.compilerOptions.types with a path to the cli d.ts file iiuc

PR #326, opened 2025-09-11
fix: allow dev command to start without entrypoint by creating defaul…

### LINKED ISSUE #325: Do not fail to start project even if the entrypoint is missing
```
No entrypoint found. Run 'tsci init' to bootstrap a basic project.
```

MAVRICK-1:
This pull request introduces an improvement to the developer experience in the CLI registration flow. Now, if no entrypoint file is found, the CLI will automatically create a default `index.tsx` file with a simple board example, instead of just prompting the user to initialize the project manually.

Enhancements to entrypoint handling:

* If no entrypoint is detected, a default `index.tsx` file is created in the current working directory with a basic board component example, making it easier for users to get started.
* User messaging has been improved: the CLI now informs the user that a default entrypoint has been created and provides a tip about using `tsci init` for a full project setup

https://github.com/user-attachments/assets/d9c1dc46-edbf-4f49-8397-7d6143f68026

/claim #325
Fixes #325

### DIFF
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -79,7 +79,7 @@ export class DevServer {
       port: this.port,
       defaultMainComponentPath: this.componentFilePath
         ? path.relative(this.projectDir, this.componentFilePath)
-        : null,
+        : undefined,
     })
     this.httpServer = server
 
--- a/cli/dev/DevServer.ts
+++ b/cli/dev/DevServer.ts
@@ -23,8 +23,9 @@ export class DevServer {
   port: number
   /**
    * The path to a component that exports a <board /> or <group /> component
+   * Can be null when running in sandbox mode without a specific file
    */
-  componentFilePath: string
+  componentFilePath: string | null
 
   projectDir: string
   /** Paths or directory names to ignore when syncing files */
@@ -57,12 +58,14 @@ export class DevServer {
     projectDir,
   }: {
     port: number
-    componentFilePath: string
+    componentFilePath?: string | null
     projectDir?: string
   }) {
     this.port = port
-    this.componentFilePath = componentFilePath
-    this.projectDir = projectDir ?? path.dirname(componentFilePath)
+    this.componentFilePath = componentFilePath || null
+    this.projectDir =
+      projectDir ??
+      (componentFilePath ? path.dirname(componentFilePath) : process.cwd())
     const projectConfig = loadProjectConfig(this.projectDir)
     this.ignoredFiles = projectConfig?.ignoredFiles ?? []
     this.fsKy = ky.create({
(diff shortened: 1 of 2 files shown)


seveibar (changes requested):
We shouldn't be creating an entrypoint, the code should just open runframe

MAVRICK-1:
> We shouldn't be creating an entrypoint, the code should just open runframe

okay understood on it

MAVRICK-1:
https://github.com/user-attachments/assets/a36f5101-492c-4ed8-9b20-f40c636cf7d2

@seveibar you are refering to this workflow ?

seveibar:
yes the user can select a file after the server starts. In my local project i have a ton of entrypoints that aren't being found and so the server doesn't start

MAVRICK-1:
https://github.com/user-attachments/assets/7b4f0d7e-33a6-4ed5-9304-d3eb72f66c02

@seveibar like this right ?

seveibar:
yep

MAVRICK-1:
@seveibar done , pushed the changes

cli/dev/register.ts
```diff
@@ -53,20 +53,30 @@ export const registerDev = (program: Command) => {
           console.log("Found entrypoint at:", entrypointPath)
         } else {
           console.log(
-            "No entrypoint found. Run 'tsci init' to bootstrap a basic project.",
+            kleur.yellow(
+              "No entrypoint found. Starting development server in sandbox mode.",
```
imrishabh18:
```suggestion
              "No entrypoint found. Starting development server.",
```

Remove all the reference of `sandbox` from the other files too! It's not a sandbox

imrishabh18 (changes requested):

MAVRICK-1:
@seveibar @imrishabh18 is it okay now?

cli/dev/DevServer.ts
```diff
@@ -271,17 +278,19 @@ export class DevServer {
...
-      onExit: () => {},
-      onError: (e) => {
-        console.error("Failed to save snippet:- ", e)
-        postEvent("FAILED_TO_SAVE_SNIPPET", e)
-      },
-      onSuccess: () => {
-        postEvent("SNIPPET_SAVED")
-      },
-    })
+    if (this.componentFilePath) {
```
seveibar: this should error instead of breaking silently

imrishabh18 (approved):
LGTM!

seveibar (changes requested):
CC @ShiboSoftwareDev after this change, automatic type installation will fail for some projects, just noting that an entrypoint is often not being found on my projects so i guess my auto type installation won't work, not a huge deal

MAVRICK-1:
@seveibar i dont know why the test are failing ? https://github.com/tscircuit/cli/pull/326/commits/1c8ba2224b038407befee2b8429794d0c70f1fa4 i have made this single change

MAVRICK-1:
@seveibar just a heads-up

PR #334, opened 2025-09-11
feat(cli): Add support for GLTF 3D models in dev server

ShiboSoftwareDev:
This PR adds support for rendering .gltf 3D models.                              

A preprocessing step has been added to the dev server. When a .gltf file is changed or loaded, the server now:                                           

 1 Reads the .gltf file and parses it as JSON.                                                                                                                     
 2 Finds all external file URIs referenced in the buffers and images properties.                                                                                   
 3 Reads the content of these local files (e.g., .bin, .png).                                                                                                      
 4 Encodes the file content into Base64 and embeds it as a data URI.                                                                                               

This process creates a self-contained .gltf file that is sent to the browser, ensuring the model and all its assets are available for rendering without requiring additional file requests. 

image


ShiboSoftwareDev:
@seveibar I still don't understand the last step
image

cli/dev/DevServer.ts
```diff
@@ -19,6 +19,37 @@ import { shouldIgnorePath } from "lib/shared/should-ignore-path"
 
 const debug = Debug("tscircuit:devserver")
 
+const processGltfFile = (absoluteFilePath: string): string => {
```
seveibar, line 22: Probably belongs in a separate file

example-dir2/gltf.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.gltf" {
+  const value: string
+  export default value
+}
```
seveibar, line 4: This should be provided by the tscircuit package or possibly the eval package, we dont want users to have to put itbin

seveibar (changes requested):
The file server should represent the operating system files- we cant keep embedding every asset.

seveibar:
Eval needs to use a “projectBaseUrl” (from the platformConfig) that can be relied upon for retrieving or importing static assets for the project

seveibar:
@ShiboSoftwareDev the last step is to generate a gltf file for a circuit that contains a gltf url (so shouldn't involve this repo)

PR #349, opened 2025-09-15
feat(dev): allow selecting entry points if missing

### LINKED ISSUE #325: Do not fail to start project even if the entrypoint is missing

ArnavK-09:
fix #325 

image

###### /claim #325


seveibar (commented):
so IMO @ArnavK-09 we should just start the server and let the default RunFrame entrypoint selection go instead of prompting the user

seveibar (approved):

PR #350, opened 2025-09-15
fix: use relative path to the directory

imrishabh18:


seveibar (changes requested):
this is in the `lib/shared`, `process.cwd()` isn't always available

seveibar:
i think the change needs to take place somewhere else? lmk what specific issue you're solving?

imrishabh18:
@seveibar

image

seveibar (approved):
sry

seveibar:
i thought this was eval, i have no idea what lib/shared refers to but it's not the browser etc.

PR #356, opened 2025-09-16
feat(spice): add simulate and export spice commands

### LINKED ISSUE #348: Command Line SPICE simulation, export to SPICE
We should be able to run spice simulations on the command line with `tsci simulate analog path/to/circuit.tsx`

We should support common ngspice outputs via `tsci export` (or whatever)

We should use the WASM build of ngspice, similar to how the schematic-viewer works https://github.com/tscircuit/schematic-viewer/blob/main/lib/workers/spice-simulation.worker.ts

If applicable, we should create a new NPM module to manage running SPICE for circuit-json with browser/node compat

CC @ShiboSoftwareDev 

/bounty $150

ShiboSoftwareDev:
This pull request introduces two new features for running circuit simulations directly from the CLI.                         

A new simulate command has been added to allow users to quickly run simulations and view the results.                        

 • Usage: tsci simulate analog <file>                                                                                        
 • Functionality:                                                                                                            
    • Runs a transient SPICE simulation on the provided .tsx file.                                                           
    • Outputs a formatted, aligned table of the simulation results directly to the console.                                  
    • Simulation is performed using the eecircuit-engine library.                                                            

The export spice commandautomatically runs a simulation and saves the results.                                          

 • Usage: tsci export <file> -f spice                                                                                        
 • Functionality:                                                                                                            
    • Exports the .spice.cir file.                                                                                           
    • Additionally, it runs a simulation and saves the results to a .csv file with the same base name.                       
    • A single console message confirms the creation of both files.                                            

Technical Details                                                                                                            

 • A runSimulation utility has been created to handle fetching, initializing, and running the eecircuit-engine WASM-based    
   simulator.                                                                                                                
 • The initialization process now uses a polling mechanism to reliably start the simulation engine.                          
 • Helper functions (resultToCsv and resultToTable) have been added to parse and format the simulation output,        
   automatically handling duplicate data columns from the engine.

simulate command:
image

/claim #348
/closes #348

### DIFF
--- a/cli/export/register.ts
+++ b/cli/export/register.ts
@@ -1,7 +1,11 @@
 import type { Command } from "commander"
 import { exportSnippet } from "lib/shared/export-snippet"
 import { generateCircuitJson } from "lib/shared/generate-circuit-json"
-import { circuitJsonToSpice } from "circuit-json-to-spice"
+import { getSpiceWithPaddedSim } from "lib/shared/get-spice-with-sim"
+import { runSimulation } from "lib/eecircuit-engine/run-simulation"
+import { resultToCsv } from "lib/shared/result-to-csv"
+import path from "node:path"
+import { promises as fs } from "node:fs"
 
 export const registerExport = (program: Command) => {
   program
@@ -16,9 +20,27 @@ export const registerExport = (program: Command) => {
       if (format === "spice") {
         const { circuitJson } = await generateCircuitJson({ filePath: file })
         if (circuitJson) {
-          const spiceNetlist = circuitJsonToSpice(circuitJson as any)
-          const spiceString = spiceNetlist.toSpiceString()
-          console.log(spiceString)
+          const spiceString = getSpiceWithPaddedSim(circuitJson as any)
+
+          const outputSpicePath =
+            options.output ??
+            path.join(
+              path.dirname(file),
+              `${path.basename(file, path.extname(file))}.spice.cir`,
+            )
+
+          await fs.writeFile(outputSpicePath, spiceString)
+
+          const { result } = await runSimulation(spiceString)
+
+          const csvContent = resultToCsv(result)
+
+          const outputCsvPath = outputSpicePath.replace(/\.spice\.cir$/, ".csv")
+
+          await fs.writeFile(outputCsvPath, csvContent)
+          console.log(
+            `Exported to ${outputSpicePath} and ${outputCsvPath} (simulation results)!`,
+          )
         }
         return
       }
--- a/cli/export/register.ts
+++ b/cli/export/register.ts
@@ -1,5 +1,7 @@
 import type { Command } from "commander"
 import { exportSnippet } from "lib/shared/export-snippet"
+import { generateCircuitJson } from "lib/shared/generate-circuit-json"
+import { circuitJsonToSpice } from "circuit-json-to-spice"
 
 export const registerExport = (program: Command) => {
   program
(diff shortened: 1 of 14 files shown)


cli/export/register.ts
```diff
@@ -9,9 +11,21 @@ export const registerExport = (program: Command) => {
...
     .option("-o, --output <path>", "Output file path")
     .action(async (file, options) => {
+      const format = options.format ?? "json"
+
+      if (format === "spice") {
+        const { circuitJson } = await generateCircuitJson({ filePath: file })
+        if (circuitJson) {
+          const spiceNetlist = circuitJsonToSpice(circuitJson as any)
+          const spiceString = spiceNetlist.toSpiceString()
+          console.log(spiceString)
```
seveibar: nice, but isn't this supposed to output to a file? I can't remember exactly. Also why are you checking for the existence of circuitJson then silently failing? If the command fails we need to be loud about it

cli/simulate/register.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const registerSimulate = (program: Command) => {
+  const simulateCommand = program
+    .command("simulate")
+    .description("Run a simulation")
+
+  simulateCommand
+    .command("analog")
+    .description("Run an analog SPICE simulation")
+    .argument("<file>", "Path to the circuit file")
```
seveibar:
```suggestion
    .argument("<file>", "Path to tscircuit tsx or circuit json file")
```

cli/simulate/register.ts
```diff
@@ -0,0 +1,36 @@
...
+        filePath: file,
+        saveToFile: false,
+      })
+      if (circuitJson) {
+        const spiceNetlist = circuitJsonToSpice(circuitJson as any)
+        let spiceString = spiceNetlist.toSpiceString()
+
+        spiceString = spiceString.replace(
+          /\.END/i,
+          ".tran 1us 1ms\n.probe V(N1) V(N2) V(N3)\n.END",
```
seveibar: this is a bit weird, this should be configured in `circuitJsonToSpice` or as part of a tscircuit element or something i.e.

in this example analogsimulation would be a group

```tsx
<analogsimulation>
  <resistor />
</analogsimulation>
```

cli/simulate/register.ts
```diff
@@ -0,0 +1,36 @@
...
+        )
+
+        const result = await runSimulation(spiceString)
+
+        console.log(JSON.stringify(result, null, 2))
```
seveibar: yea i think the user should be able to configure outputs, json isn't bad but CSV is a bit more conventional with ngspice i think

cli/simulate/register.ts
```diff
@@ -0,0 +1,36 @@
...
+      } else {
+        console.log("error: Failed to generate circuit JSON")
```
seveibar: use early bail, don't have big if elses

lib/eecircuit-engine/run-simulation.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const fetchSimulation = async (): Promise<
+  typeof EecircuitEngine.Simulation
+> => {
+  // Using a predictable path in tmp to avoid re-downloading.
+  // This is a workaround for Bun's test environment not handling URL imports.
+  const tempFilePath = path.join(os.tmpdir(), "eecircuit-engine-1.5.2.mjs")
+
+  if (!existsSync(tempFilePath)) {
+    const url = "https://cdn.jsdelivr.net/npm/eecircuit-engine@1.5.2/+esm"
```
seveibar, line 22: I think we can bundle this inside the CLI since the CLI isn't used in browser- i don't mind this approach though, and actually if we save it to e.g. `${USER_CONFIG_DIR}/.tscircuit/eecircuit-engine-build.js` or something it could be nice to dynamically load some of these dependencies, as long as we don't dynamically load them every time

Don't worry about this too much, i think we can adjust this in the future. It's a nice implementation, just has some loading cost

cli/simulate/register.ts
ShiboSoftwareDev: @seveibar this data is a requirement to run the simulation, are you saying whenever we want to have a simulation users would be required to use `analogsimulation` or such components??

seveibar (approved):

seveibar (approved):
AWESOME

PR #435, opened 2025-10-03
    switch 3d PNG generation to poppygl from simple-3d for the `tscircuit build`

imrishabh18:
image


seveibar (approved):
Yes we can also remove simple 3d svg fwiw

PR #511, opened 2025-10-09
feat: kicad import example & sort examples directories

ArnavK-09:
image


lib/site/getIndex.ts
```diff
@@ -1,17 +1,23 @@
...
+export const getIndex = async (
+  mainComponentPath?: string,
+  fileServerApiBaseUrl?: string,
+) => {
   const sessionToken = getSessionToken()
   const tokenScript = sessionToken
     ? `\n        window.TSCIRCUIT_REGISTRY_TOKEN = ${JSON.stringify(sessionToken)};`
     : ""
+  const fileServerApiScript = fileServerApiBaseUrl
+    ? `\n        window.TSCIRCUIT_FILESERVER_API_BASE_URL = ${JSON.stringify(fileServerApiBaseUrl)};`
```
seveibar, line 12: haha i know what you're hacking here and I don't like it. Don't avoid adding support for `/...` style urls!

seveibar (commented):
i don't think we need the window var, i think i know what you're trying to hack but it's not the right way!

The example is good though! i just don't like the window var being added (shouldn't be necessary)

example-dir2/kicad_footprint.circuit.tsx
```diff
@@ -0,0 +1,15 @@
+// @ts-ignore
```
seveibar: don't pollute the top level directories. Either re-use the example-dir or make a directory `examples`

seveibar (commented):
we _must_ fix the `new URL` bug, by removing the reproduction you're creating an even harder bug to solve

seveibar (approved):

seveibar (approved):

PR #525, opened 2025-10-10
GLB Loading Examples (reproduction for project-relative-path loading issue)

seveibar:


examples/glb-loading/glb-project-relative-path.tsx
```diff
@@ -0,0 +1,17 @@
+export default () => {
+  return (
+    <board width="12mm" height="30mm">
+      <chip
+        name="H1"
+        pcbX={0}
+        pcbY={0}
+        cadModel={<cadmodel modelUrl="/examples/glb-loading/soic8.glb" />}
```
seveibar, line 8: @ArnavK-09 this triggers the bug, if you run `bun run start` you can see it locally

PR #784, opened 2025-10-30
feat: add favicon link to HTML template in getIndex function

ArnavK-09:


seveibar (approved):
might be better to link to the tscircuit.com version

ArnavK-09:
> might be better to link to the tscircuit.com version

it's url is randomized each build

PR #828, opened 2025-11-04
Add support for `tsci build --transpile`

imrishabh18:
image

/close #819

PR #841, opened 2025-11-05
feat: add support for transpilation `tsci build --transpile`

imrishabh18:
Same ESM output

image

image

#828

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "devDependencies": {
     "@babel/standalone": "^7.26.9",
     "@biomejs/biome": "^1.9.4",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -2,6 +2,13 @@
   "name": "@tscircuit/cli",
   "version": "0.1.446",
   "main": "dist/main.js",
+  "dependencies": {
```
imrishabh18, line 5: Either we can add the rollup inside the `dependencies` and not `devDeps` or in the tscircuit package because they need their native binaries available at runtime.

@seveibar

seveibar (approved):

PR #877, opened 2025-11-08
fix: improve path handling in build

ArnavK-09:
before vs after

image

### DIFF
--- a/lib/shared/generate-circuit-json.tsx
+++ b/lib/shared/generate-circuit-json.tsx
@@ -1,7 +1,7 @@
 import { getVirtualFileSystemFromDirPath } from "make-vfs"
-import path from "node:path/posix"
-import { relative } from "node:path"
+import path from "node:path"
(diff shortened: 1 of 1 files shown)


lib/shared/generate-circuit-json.tsx
```diff
@@ -48,15 +48,18 @@ export async function generateCircuitJson({
   const runner = new userLandTscircuit.RootCircuit({
     platform: platformConfig,
   })
-  const projectDir = path.dirname(filePath)
-  const resolvedOutputDir = outputDir || projectDir
+  const absoluteFilePath = path.isAbsolute(filePath)
```
imrishabh18, line 51:
```suggestion
  const resolvedFilePath = path.isAbsolute(filePath)
```

This can be absolute or relative right?

ArnavK-09: yes
image
image

PR #909, opened 2025-11-11
use CircuitRunner from eval for generating circuit json (supports fetching external packages from registry)

imrishabh18:
The `tsci build` fails for the circuit which has `import {} from "@tsci/something"`


seveibar (changes requested):
we should not be using fsmap here

seveibar:
that error just means you need to install the package

PR #923, opened 2025-11-12
feat: Add the support for the install command `tsci install`

imrishabh18:
image

### DIFF
--- a/lib/shared/collect-tsci-dependencies.ts
+++ b/lib/shared/collect-tsci-dependencies.ts
@@ -2,9 +2,7 @@ import fs from "node:fs"
 import path from "node:path"
 import { globbySync } from "globby"
 
-const DEFAULT_PATTERNS = [
-  "**/*.{ts,tsx,js,jsx}",
-]
+const DEFAULT_PATTERNS = ["**/*.{ts,tsx,js,jsx}"]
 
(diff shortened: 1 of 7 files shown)


lib/shared/generate-circuit-json.tsx
```diff
@@ -43,6 +43,9 @@ export async function generateCircuitJson({
 }: GenerateCircuitJsonOptions) {
   debug(`Generating circuit JSON for ${filePath}`)
 
+  // Import React and make it globally available for packages referencing it
+  const React = await importFromUserLand("react")
```
imrishabh18, line 47: image

I get this error if we don't add `react` globally

lib/shared/collect-tsci-dependencies.ts
```diff
@@ -0,0 +1,56 @@
+import fs from "node:fs"
+import path from "node:path"
+import { globbySync } from "globby"
+
+const DEFAULT_PATTERNS = [
+  "**/*.circuit.{ts,tsx,js,jsx}",
+  "**/*.snippet.{ts,tsx,js,jsx}",
```
seveibar: Why not look at all tsx files?

seveibar (approved):

PR #954, opened 2025-11-17
feat: support local packages to be used by circuit on `tsci dev`

imrishabh18:
image

### DIFF
--- a/lib/dependency-analysis/getNodeModuleDependencies.ts
+++ b/lib/dependency-analysis/getNodeModuleDependencies.ts
@@ -1,0 +1,331 @@
+import * as ts from "typescript"
+import * as path from "path"
+import * as fs from "fs"
+
+interface NodeModuleDependency {
+  /** The package name (e.g., "react" or "@tscircuit/core") */
+  packageName: string
+  /** The import path used in the source (e.g., "react/jsx-runtime") */
+  importPath: string
+  /** Resolved absolute file paths for this dependency */
+  resolvedFiles: string[]
+}
+
+/**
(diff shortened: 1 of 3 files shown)


lib/dependency-analysis/getNodeModuleDependencies.ts
```diff
@@ -0,0 +1,423 @@
...
+      path.join(packageDir, "index.js"),
+      path.join(packageDir, "index.mjs"),
+      path.join(packageDir, "index.ts"),
+      path.join(packageDir, "index.tsx"),
+      path.join(packageDir, "dist", "index.js"),
+      path.join(packageDir, "dist", "index.mjs"),
+      path.join(packageDir, "lib", "index.js"),
+      path.join(packageDir, "src", "index.ts"),
+      path.join(packageDir, "src", "index.tsx"),
+    ]
```
seveibar: don't bother with the fallbacks, it wouldn't work with `tsci build` or any node runtime if `main` isn't defined

lib/dependency-analysis/getNodeModuleDependencies.ts
```diff
@@ -0,0 +1,423 @@
...
+      }
+    }
+  }
+
+  // Main entry points from package.json
+  const entryPoints = [
+    packageJson.main,
+    packageJson.module,
+    packageJson.types,
+    packageJson.typings,
```
seveibar:
```suggestion
```

seveibar (approved):
lg, it is possible we will need to add support for tsconfig aliases when evaluating if something is a node module or not

PR #1002, opened 2025-11-21
TSC-112 Support installing espressif kicad libraries using tsci install

MustafaMulla29:


cli/install/register.ts
```diff
@@ -1,15 +1,63 @@
...
+
+            console.log(`\n✓ Successfully installed ${packageArg}`)
+          } else {
+            // Install as a regular npm package
+            packageManager.install({ name: packageArg, cwd })
+            console.log(`\n✓ Successfully installed ${packageArg}`)
+          }
+        } else {
+          // No package specified, install all dependencies
+          await installProjectDependencies({ cwd })
```
seveibar: long if elses are an anti-pattern, move this to the top and do early return

cli/install/register.ts
```diff
@@ -1,15 +1,63 @@
...
+            await generateKicadTypeDeclarations(cwd)
+
+            // Show usage examples
+            await generateKicadExports(packageArg, cwd)
+
+            console.log(`\n✓ Successfully installed ${packageArg}`)
+          } else {
+            // Install as a regular npm package
+            packageManager.install({ name: packageArg, cwd })
+            console.log(`\n✓ Successfully installed ${packageArg}`)
```
seveibar: again, this is a long if else antipattern. The logic itself is fine, but then you also have repeated console.logs

lib/shared/install-kicad-library.ts
```diff
@@ -0,0 +1,149 @@
...
+    return { owner, repo: repo.replace(/\.git$/, "") }
+  }
+
+  return null
+}
+
+/**
+ * Generates TypeScript declarations for .kicad_mod files in types/ folder
+ */
+export async function generateKicadTypeDeclarations(
```
seveibar: each of these should be in it's own file, with it's function as the file name

lib/shared/install-kicad-library.ts
```diff
@@ -0,0 +1,149 @@
...
+  // Create types directory if it doesn't exist
+  const typesDir = path.join(cwd, "types")
+  if (!fs.existsSync(typesDir)) {
+    fs.mkdirSync(typesDir, { recursive: true })
+  }
+
+  // Generate type declaration file in types/ folder
+  const typeDeclarationPath = path.join(typesDir, "kicad_mod.d.ts")
+  const typeDeclaration = `declare module "*.kicad_mod" {
+  const value: string
```
seveibar: it loads as a circuit json array or any[] IIRC, you might be right but would appreciate double checking

seveibar (changes requested):
very good but see comments

cli/install/register.ts
```diff
@@ -1,15 +1,43 @@
...
+
+        const packageManager = getPackageManager()
+
+        // Check if it's a GitHub URL (for KiCad libraries, etc.)
+        if (!isGitHubUrl(packageArg)) {
+          // Install as a regular npm package
+          packageManager.install({ name: packageArg, cwd })
+        } else {
+          // Handle GitHub repository installation
+          await installKicadLibrary(packageArg, cwd)
```
seveibar: just because it's a github url, doesn't mean it's a kicad library, change this name to `installGithubLibrary`

seveibar: also this logic is really silly, `if (isGithubUrl(...)) { // do github url stuff }`

flip this if condition so that you're handling github installation when you check for github urls, the default behaviort is to install a package, not to assume it's a github url

lib/shared/generate-kicad-exports.ts
```diff
@@ -0,0 +1,63 @@
+import fs from "node:fs"
+import path from "node:path"
+import { glob } from "glob"
+import { extractGitHubInfo } from "./extract-github-info"
+
+/**
+ * Shows usage examples for imported KiCad footprints
+ */
+export async function generateKicadExports(
```
seveibar: function name doesn't match what it does.

You're supposed to generate a file `types/${module}.d.ts` with all the imports of the package. Be explicit. `generateKicadRepoTypeDeclarations`. Remove the logging from this function- it could go somewhere else but it doesn't belong in a function with either of these names

seveibar: i'm reading more and it seems like you already have that function- so rename this function to printKicadRepoUsage`

lib/kicad/generate-kicad-type-declarations.ts
```diff
@@ -0,0 +1,47 @@
...
+  const value: AnyCircuitElement[]
+  export default value
+}
+`
+
+  fs.writeFileSync(typeDeclarationPath, typeDeclaration)
+  console.log(
+    `Generated type declarations for ${kicadModFiles.length} .kicad_mod file(s)`,
+  )
+  console.log(`Type declarations saved to: types/kicad_mod.d.ts`)
```
seveibar: ${module}.d.ts, include a snapshot of this file in your test using toMatchInlineSnapshot

lib/shared/install-github-library.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getPackageManager } from "./get-package-manager"
+
+/**
+ * Installs a KiCad library from a GitHub repository
+ */
+export async function installKicadLibrary(
+  packageArg: string,
+  cwd: string = process.cwd(),
+): Promise<void> {
+  console.log(`Detected GitHub repository`)
```
seveibar, line 13: improve naming.......

tests/cli/install/install-kicad-library.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    expect(existsSync(typesPath)).toBe(true)
+
+    const typesContent = readFileSync(typesPath, "utf-8")
+    expect(typesContent).toContain("declare module")
+    expect(typesContent).toContain("*.kicad_mod")
+
+    // Verify bun install reinstalls the library
+    const { $ } = Bun
+    await $`rm -rf ${join(tmpDir, "node_modules")}`.cwd(tmpDir).quiet()
+    expect(existsSync(nodeModulesPath)).toBe(false)
```
seveibar: delete all of the assertions you have here, it's all ai slop

test for the things that actually matter. Are you able to use the code as you described? What is logged? What does the types.d.ts file look like?

Seriously, put time into demonstrating what the code does and that it works. Write it by hand, do not use AI

seveibar (changes requested):
- Put your kicad specific functionality into `lib/kicad`
- **Dramatically improve the test, please do not generate a bunch of assertions. Use the imports and verify that your code actually works, your test does not actually test usage, what the types.d.ts file looks like**

PR #1013, opened 2025-11-22
Support installing espressif kicad libraries using tsci install

MustafaMulla29:


MustafaMulla29:
bun test failing coz of the toMatchInlineSnapshot in the test

PR #1016, opened 2025-11-22
Support tsci intall "github url" command TSC-112 

MustafaMulla29:
image

Runframe/dev: Sets projectBaseUrl → core generates proper URLs → fetch works
CLI build: No projectBaseUrl → core passes raw paths → we convert to  for Bun's fetch

In our CLI build case, we're not setting a projectBaseUrl, so @tscircuit/core just passes the raw file path. That's why we need to convert it to a URL, solution is to pass projectBaseUrl: "file://" in the config.


cli/install/register.ts
```diff
@@ -1,15 +1,22 @@
 import type { Command } from "commander"
 import { installProjectDependencies } from "lib/shared/install-project-dependencies"
+import { installGithubKicadLibrary } from "../../lib/shared/install-github-library"
 
 export const registerInstall = (program: Command) => {
   program
-    .command("install")
+    .command("install [githubUrl]")
```
seveibar: oh hmm, i guess `tsci add` should also support this 🤔 , and it should be `packageOrUrl`

cli/install/register.ts
```diff
@@ -1,15 +1,22 @@
...
-      "Install project dependencies and generate package.json if needed",
+      "Install project dependencies or install a GitHub KiCad library (e.g., tsci install https://github.com/espressif/kicad-libraries)",
     )
-    .action(async () => {
+    .action(async (githubUrl?: string) => {
       try {
-        await installProjectDependencies()
+        if (githubUrl) {
+          // Install GitHub KiCad library
+          await installGithubKicadLibrary(githubUrl)
```
seveibar: yea i'm not sure why you have `githubUrl` as a var here, it's any valid `npm add` or `npm install` target, we use the same parameters as npm, this is the same as `bun add githubUrl`

lib/kicad/generate-types.ts
```diff
@@ -0,0 +1,126 @@
...
+    console.log(kleur.gray("Creating tsconfig.json..."))
+    const tsconfig = {
+      compilerOptions: {
+        target: "ES2020",
+        module: "ESNext",
+        moduleResolution: "bundler",
+        jsx: "react",
+        jsxImportSource: "@tscircuit/core",
+        types: ["tscircuit"],
+        typeRoots: ["./node_modules/@types", "./types"],
```
seveibar: you should be loading this from the same source as `tsci init`

lib/shared/generate-circuit-json.tsx
```diff
@@ -48,8 +48,26 @@ export async function generateCircuitJson({
   ;(globalThis as any).React = React
   const userLandTscircuit = await importFromUserLand("tscircuit")
 
+  // Get default platform config with KiCad parsing support
+  const { getPlatformConfig } = await import("@tscircuit/eval")
```
seveibar: you should not modify the platformConfig that's passed into this function, the platformConfig should be modified in a different place

cli/install/register.ts
```diff
@@ -1,15 +1,22 @@
...
-      "Install project dependencies and generate package.json if needed",
+      "Install project dependencies or install a KiCad library package (e.g., tsci install https://github.com/espressif/kicad-libraries or tsci install kicad-libraries@1.0.0)",
     )
-    .action(async () => {
+    .action(async (packageSpec?: string) => {
       try {
-        await installProjectDependencies()
+        if (packageSpec) {
+          // Install KiCad library package
+          await installKicadLibrary(packageSpec)
```
seveibar, line 15: Cant believe im still saying this but this code doesnt make sense. You dont know its a kicad library yet! At this point you should be chrcking what the url is etc.

Just revert this file, focus on tsci add

PR #1031, opened 2025-11-24
fix: transpilation output files with assets directory as well as correct esm and cjs output

imrishabh18:


cli/build/transpile/index.ts
```diff
@@ -15,16 +15,36 @@ import {
...
+    const potentialPaths = [
+      path.join(projectDir, id),
+      path.join(projectDir, `${id}.ts`),
+      path.join(projectDir, `${id}.tsx`),
+      path.join(projectDir, `${id}.js`),
+      path.join(projectDir, `${id}.jsx`),
+      path.join(projectDir, id, "index.ts"),
+      path.join(projectDir, id, "index.tsx"),
+      path.join(projectDir, id, "index.js"),
+      path.join(projectDir, id, "index.jsx"),
```
seveibar: shouldn't we be using tsconfig paths?

cli/build/transpile/index.ts
```diff
@@ -68,27 +125,31 @@ export const transpileFile = async ({
...
           declaration: false,
           sourceMap: false,
           skipLibCheck: true,
           resolveJsonModule: true,
           allowSyntheticDefaultImports: true,
           allowArbitraryExtensions: true,
+          baseUrl: projectDir,
           ...(typeRoots.length ? { typeRoots } : {}),
+          ...(validTypes.length ? { types: validTypes } : {}),
         },
```
seveibar: hmm interesting, why aren't we loading the users tsconfig?

seveibar (approved):

PR #1071, opened 2025-11-26
fix: replace the bundled static file with varialbe to an import statement

imrishabh18:
Eval resolves the path for those statement which are having `import`, and the transpiled path has path in the variable due to which it was not getting resolved in the eval

image

image

image


seveibar (approved):

seveibar (changes requested):
assets need to be imported in the `dist` files- we need proof of that inside the test, e.g. with toMatchInlineSnapshot against a dist/index.js file

PR #1148, opened 2025-12-01
Support kicad installation for 'tsci install' command and add colors using kleur

MustafaMulla29:


tests/cli/install/install-kicad-library.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    expect(existsSync(typesFilePath)).toBe(true)
+
+    console.log("✓ Types file generated")
+
+    // Verify types content
+    const typesContent = readFileSync(typesFilePath, "utf-8")
+    expect(typesContent).toContain('declare module "kicad-libraries/')
+    expect(typesContent).toContain(".kicad_mod")
+    expect(typesContent).toContain("const value: string")
+    expect(typesContent).toContain("export default value")
```
seveibar, line 73: this is hard to read, just use `toMatchInlineSnapshot` for the file content

seveibar (approved):

tests/cli/install/install-kicad-library.test.ts
MustafaMulla29: The output is very large in it, so using this only instead of toMatchInlineSnapshot

seveibar: why is the output large?

PR #1240, opened 2025-12-03
Fix kicad_mod files not uploading

MustafaMulla29:
image


seveibar (approved):

seveibar:
should really have a test for this!!!! otherwise someone WILL break it!

PR #1251, opened 2025-12-04
Don't install type dependencies for snippets, since they are included inside node_modules

MustafaMulla29:
Fixes TSC-191


imrishabh18 (commented):
You can get rid of the file `installNodeModuleTypesForSnippet` as we don't need it anymore. Also, Can you verify on your machine that after installing a new package while the devServer is running you don't get type errors in the editor and the types are loaded from the `node_modules`

MustafaMulla29:
@imrishabh18 Works! Types are bundled in node_modules. And no type errors in the editor

seveibar (changes requested):
just remove the tests, we don't need to test negatives (just adds bloat)

PR #1287, opened 2025-12-06
Support kicad-symbols-library for the tsci export command

MustafaMulla29:
Folder after unzipping:
image


seveibar:
Why would we support this? This is just exporting tscircuit builtin symbols- until custom symbol support is working it's difficult to know if this works

seveibar:
i'm not sure in what scenario you'd want to export kicad symbols from a tscircuit project- the time that could make sense is if you're also exporting a kicad project (and it's included inside)

MustafaMulla29:
> i'm not sure in what scenario you'd want to export kicad symbols from a tscircuit project- the time that could make sense is if you're also exporting a kicad project (and it's included inside)

Inside a kicad-footprint-library?

PR #1309, opened 2025-12-10
feat: add package name validation to pushSnippet using `tsci push`

Ayushjhawar8:
Before: 
image
After:
image


lib/shared/push-snippet.ts
```diff
@@ -89,26 +90,68 @@ export const pushSnippet = async ({
...
     }
 
-    if (unscopedPackageName.includes("/")) {
+    if (inputName.includes("/")) {
       onError("Package name cannot contain a '/'")
       return onExit(1)
     }
 
+    // Sanitize the package name
+    unscopedPackageName = sanitizePackageName(inputName)
```
Ayushjhawar8: this is when package.json doesn't have a name and the user is prompted to enter one

lib/shared/push-snippet.ts
```diff
@@ -89,26 +90,68 @@ export const pushSnippet = async ({
   let unscopedPackageName = getUnscopedPackageName(packageJson.name ?? "")
   const packageJsonAuthor = getPackageAuthor(packageJson.name ?? "")
 
+  // Sanitize the unscoped package name if it already exists
+  if (unscopedPackageName) {
```
Ayushjhawar8, line 94: this is when package.json already has a name (e.g., "My Test Package With Spaces")

seveibar (changes requested):
the code is correct as-is. We should error not change the package name

Ayushjhawar8:
image
image

seveibar (approved):

PR #1321, opened 2025-12-11
fix: improve test reliability with retry logic and increase timeout

Ayushjhawar8:
1. `scripts/run-tests-with-retry.ts`  (new file)
- Increases per-test timeout: 20s → 30s
- Retries only the specific failed tests (up to 3 attempts)
- Preserves existing segfault retry logic (up to 4 attempts)
- Cross-platform compatible using  `path.basename()`
2.  `.github/workflows/bun-test.yml`
- Workflow timeout: 15 → 20 minutes
- Uses helper script instead of inline logic
3.  `.github/workflows/windows-tests.yml`
- Workflow timeout: 10 → 15 minutes
- Uses helper script instead of inline logic

### DIFF
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -9,7 +9,7 @@ jobs:
   windows-tests:
     runs-on: windows-latest
     if: "${{ github.event_name != 'pull_request' || github.event.pull_request.title != 'chore: bump version' }}"
-    timeout-minutes: 10
+    timeout-minutes: 15
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


.github/workflows/windows-tests.yml
```diff
@@ -24,4 +24,7 @@ jobs:
         run: bun install
 
       - name: Run Windows-only tests
-        run: bun test tests/windows --timeout 20_000
\ No newline at end of file
+        shell: bash
+        run: |
+          bun run scripts/run-tests-with-retry.ts --timeout 30000 tests/windows/*.test.ts
```
seveibar: !!!! this is totally different! You can't change the glob!

seveibar: tests/windows/**/*.test.ts would be an actual glob

seveibar (approved):
we can try this but you must fix and carefully check the windows tests!!!

after merging you are responsible if tests are not running on CI- so please make sure that tests aren't being skipped- if we merged this change with the windows glob you had inserted tests would not have run- please double check!!!!! If tests don't run, people will begin to break things in future PRs and we won't know!

PR #1326, opened 2025-12-12
feat: enhance login action to fetch and display GitHub username from config

rushabhcodes:
This pull request improves the login experience in the CLI by displaying the logged-in GitHub username when a user is already authenticated. It fetches the username from configuration or the registry API, with special handling for test environments.

Enhancements to login feedback:

* Updated the login action in `register.ts` to display the GitHub username of the currently logged-in user by fetching it from the config or the registry API, including fallback logic for test mode.

https://linear.app/tscircuit/issue/TSC-230/tsci-login-should-print-the-account-name-of-the-logged-in-user-if

### DIFF
--- a/cli/auth/login/register.ts
+++ b/cli/auth/login/register.ts
@@ -35,7 +35,8 @@ export const registerAuthLogin = (program: Command) => {
         account = await tryFetchAccount(githubUsernameFromConfig)
       }
 
-      const tscircuitHandle = account?.tscircuit_handle ?? account?.github_username ?? "(unknown)"
+      const tscircuitHandle =
+        account?.tscircuit_handle ?? account?.github_username ?? "(unknown)"
 
       console.log(
         `Already logged in as ${tscircuitHandle}! Use 'tsci logout' if you need to switch accounts.`,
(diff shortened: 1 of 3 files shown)


imrishabh18 (changes requested):
You need to print the tscircuit handle and not the github username

rushabhcodes:
@imrishabh18 do we also need to update `whoami` then
image

cli/auth/login/register.ts
```diff
@@ -1,16 +1,43 @@
...
+          try {
+            const { account } = await ky
+              .post<EndpointResponse["accounts/get"]>("accounts/get", {
+                json: { account_id: accountId },
+              })
+              .json()
+            return account
+          } catch {
+            return undefined
+          }
```
imrishabh18: Don't silent the errors!!

tests/cli/auth/login.test.ts
```diff
@@ -20,8 +20,8 @@ test("login command: already logged in", async () => {
 
   // TODO: Remove this when the autorouter is not emitting this warning
   expect(stderr).toBe("")
-  expect(stdout).toContain(
-    "Already logged in! Use 'tsci logout' if you need to switch accounts.",
+  expect(stdout).toMatch(
+    /Already logged in as .*! Use 'tsci logout' if you need to switch accounts\./,
```
imrishabh18: It should work without the regex too! just update with the tscircuitHandle `test-user` or something

imrishabh18 (changes requested):

rushabhcodes:
It passes locally
image

PR #1331, opened 2025-12-12
chore: update circuit-json to v0.0.333

techmannih:


seveibar (approved):
this dependency should be synced with core like tscircuit/tscircuit

PR #1481, opened 2025-12-22
feat: TSC-212 add a static dev mode for viewing Circuit JSON build artifacts

Ayushjhawar8:
…and enhance dev command

Before: 
image

After:
image

### DIFF
--- a/cli/dev/register.ts
+++ b/cli/dev/register.ts
@@ -9,6 +9,10 @@ import { getEntrypoint } from "lib/shared/get-entrypoint"
 import { globbySync } from "globby"
 import { findBoardFiles } from "lib/shared/find-board-files"
 import { DEFAULT_IGNORED_PATTERNS } from "lib/shared/should-ignore-path"
+import {
+  getStaticBuildInfoFromPath,
+  startStaticFileServer,
+} from "./static-file-server"
 
 const findSelectableTsxFiles = (projectDir: string): string[] => {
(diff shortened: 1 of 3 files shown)


imrishabh18 (changes requested):
With this change passing a directory could unexpectedly trigger the static mode. We should have some flag `--static` specifically for this

cc @seveibar

cli/dev/register.ts
```diff
@@ -83,8 +87,27 @@ export const registerDev = (program: Command) => {
...
+            files: staticBuildInfo.files,
+            startTime,
+          })
+          return
+        }
+
         if (!absolutePath.endsWith(".tsx") && !absolutePath.endsWith(".ts")) {
-          console.error("Error: Only .tsx files are supported")
+          console.error(
+            "Error: Only .tsx/.ts files or Circuit JSON files are supported",
```
seveibar, line 109:
```suggestion
            "Error: Only .tsx/.ts files or .circuit.json files are supported",
```

seveibar (commented):
so the implementation i was thinking of you would still run the regular server, but it would just be able to handle circuit.json files now
another way to put it- tsci dev should allow viewing of circuit.json files
i'm not sure much of a cli change is needed, seems more like a runframe thing? There is the tsci dev <dir> aspect

Ayushjhawar8:
Unstale

PR #1503, opened 2025-12-25
switch to tscirucit handle to push pkgs

ArnavK-09:


lib/shared/push-snippet.ts
```diff
@@ -44,6 +45,8 @@ export const pushSnippet = async ({
     return onExit(1)
   }
 
+  const account = await fetchAccount()
```
seveibar: you shouldn't need to fetch the account to get the tscircuit_handle

seveibar (approved):
see comment, please avoid refetching account because it will create issues with error management

seveibar (approved):

PR #1517, opened 2025-12-29
Add the flag `--use-cdn-javascript` to use the cdn url for the script src

imrishabh18:
For `tscircuit.app` we will be needing the site `index.html` be having the cdn url rather than the relative `standalone.min.js` file reference


seveibar (changes requested):
maybe this should be a flag

seveibar:
`--use-cdn-javascript`?

PR #1545, opened 2026-01-05
use tscircuit handle to get org

ArnavK-09:


seveibar (approved):

tests/cli/export/export-kicad.test.ts
imrishabh18, line 1: Don't add the skip! These are not failing locally right. Might be some issue with the CI

PR #1563, opened 2026-01-06
Auto-render named exports for kicad-library

MustafaMulla29:
**Before**
image

**After**
image

requires: https://github.com/tscircuit/circuit-json-to-kicad/pull/45

### DIFF
--- a/lib/shared/generate-circuit-json.tsx
+++ b/lib/shared/generate-circuit-json.tsx
@@ -107,7 +107,7 @@ export async function generateCircuitJson({
     runner.add(<MainComponent.default />)
   } else {
     // No default export: render all named component exports
-    // This is library mode - each named export becomes a component with its export name
+    // Each named export is rendered directly (must be a board-level component or wrapped in a board)
     const componentExports = Object.entries(MainComponent).filter(
       ([name, value]) =>
         name[0] === name[0].toUpperCase() && typeof value === "function",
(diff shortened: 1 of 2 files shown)


MustafaMulla29:
@seveibar is it hacky?

lib/shared/generate-circuit-json.tsx
```diff
@@ -102,25 +102,45 @@ export async function generateCircuitJson({
...
+        `Library mode: rendering ${componentExports.length} named exports: ${componentExports.map(([name]) => name).join(", ")}`,
+      )
+
+      const LibraryBoard = () => (
+        <board>
+          {componentExports.map(([exportName, Component]: [string, any], i) => (
+            <Component key={exportName} name={exportName} pcbX={i * 10} />
+          ))}
+        </board>
+      )
```
imrishabh18: This is definitely a hack!!

imrishabh18 (changes requested):

lib/shared/generate-circuit-json.tsx
imrishabh18: It will generate incorrect circuit json

MustafaMulla29: Yes that's what I was concerned with, what do you think where could be the solution?

imrishabh18: Maybe generate individual circuit json for each of those files.

cc @seveibar

seveibar (changes requested):
You should be genrating an fsmap from your kicad converters

PR #1611, opened 2026-01-14
fix: add React dependencies to transpile tests

Ayushjhawar8:
fix: https://github.com/tscircuit/cli/actions/runs/20983589977/job/60327244992


imrishabh18 (changes requested):
I don't know what you are trying to do by adding react, and why did you change the tests. The error message is trying to say that the path of the file with the `--transpile` flag doesn't work so look into that

PR #1613, opened 2026-01-14
throw error if all circuit file fails

ArnavK-09:


seveibar (changes requested):
the logic is as follows:

- if circuit json isn't produced, we call that a fatal error
- we fail if there are ANY fatal errors. fatal errors are non-recoverable
- hasErrors is not fully descriptive. there are DRC errors and warnings that will be often ignored

PR #1614, opened 2026-01-14
Remove the stringification of the err and use the native error formatting of bun with line number and file location

imrishabh18:
image

the step files are being parsed as javascript


seveibar (approved):
hmm interesting, so the step files as binary are causing issues here too

imrishabh18:
tests failing are related to the transpile path issue, merging it without those being passed

PR #1627, opened 2026-01-15
fix: static asset file imports which are present in the node_module pacakges are parsed as JS

imrishabh18:
Fix this syntax error of parsing staticAssets as javascript

image

### DIFF
--- a/cli/main.ts
+++ b/cli/main.ts
@@ -30,6 +30,10 @@ import { registerSimulate } from "./simulate/register"
 import { registerInstall } from "./install/register"
 import { registerTranspile } from "./transpile/register"
 
+// Register static asset loader plugin to handle .glb, .step, etc. imports
+// from node_modules packages that contain binary asset imports
+import "../lib/shared/static-asset-loader"
+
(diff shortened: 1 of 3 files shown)


seveibar (commented):
hold on i'm confused, i merged a similar fix but fixed some bugs (or maybe i didn't merge...)

seveibar:
https://github.com/tscircuit/cli/pull/1623

seveibar:
@imrishabh18 it looks like those STEP files are stored incorrectly- are you using an old repro? there was a bug where STEP files were being stored as binary that SHOULD be resolved now- but if the STEP files are stored as not-text they can't be loaded

imrishabh18:
> @imrishabh18 it looks like those STEP files are stored incorrectly- are you using an old repro? there was a bug where STEP files were being stored as binary that SHOULD be resolved now- but if the STEP files are stored as not-text they can't be loaded

I see, I was testing the old repro for the scaffold package. Anyways, the package build is still failing though taking a look into that

PR #1656, opened 2026-01-19
use eval to build

ArnavK-09:
#1657


seveibar (changes requested):
don't use eval to build in this module, native bun imports are used intentionally

PR #1658, opened 2026-01-19
Support `--concurrency` to build indivudual circuit files concurrently

imrishabh18:


seveibar (commented):
can you run a performance test? This will not use multiple threads

PR #1665, opened 2026-01-20
feat!: add `init` command to scaffold basic files & claude skill

ArnavK-09:


examples/basic2/tscircuit.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://cdn.jsdelivr.net/npm/@tscircuit/cli/types/tscircuit.config.schema.json",
+  "mainEntrypoint": "index.tsx"
```
seveibar:
```suggestion
```

lib/shared/generate-claude-md.ts
```diff
@@ -0,0 +1,139 @@
...
+- Capacitance: \`"100nF"\`, \`"10uF"\`, \`"1pF"\`
+- Inductance: \`"10uH"\`, \`"100nH"\`
+
+## Before You're Done
+
+- Run \`bun run build\` to verify the circuit compiles
+- Run \`bun run snapshot\` to check visual output
+`
+  writeFileIfNotExists(claudeMdPath, claudeMdContent)
+}
```
seveibar: Can you not install `CLAUDE.md` and instead install this skill somehow? And always clone the latest version?

https://github.com/tscircuit/skill

We could also just prompt the user to set up the skill

ArnavK-09:
Ok fetching skill live

lib/shared/generate-readme-md.ts
```diff
@@ -0,0 +1,73 @@
...
+
+Push to the tscircuit registry:
+
+\`\`\`bash
+tsci push
+\`\`\`
+
+## License
+
+MIT
```
seveibar: oof definitely don't add a license for a user!!!!

Do not add a README for a user!!!! We do AI-generated READMEs on tscircuit.com when users don't define them and I think that is a great default behavior (no readme = generated readme)

lib/shared/setup-github-actions.ts
```diff
@@ -50,12 +52,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 22
       - uses: oven-sh/setup-bun@v2
       - run: bun install
-      - run: bunx tsci snapshot
+      - run: bun x tsci snapshot
```
seveibar: i'm surprised we had this- you can flag-gate it if you want but i don't think we'll support workflows now that tscircuit.com is becoming functional (plus it will cause people to not use tscircuit.com if they have build in github, they'll have to struggle with github artifacts/publishing, not as good of an experience)

seveibar (changes requested):
clone the claude skill. copying that code over is a bad idea for keeping in sync

seveibar (approved):

PR #1675, opened 2026-01-20
fix: recusion of `prebuildCommand` running infinitely if the `buildCommand` having the `--ci` flag

imrishabh18:
image


seveibar (approved):

seveibar:
this is the wrong way to fix this, i will describe the right way shortly

seveibar:
@imrishabh18 have you tried using `import.meta.main`, that should be `true` if it is the main file path

PR #1751, opened 2026-01-23
Convert author and packageName to lowercase for kicadPcmPackageId

MustafaMulla29:


MustafaMulla29:
Not required rn, as kicad resolves it to lowercase automatically

PR #1765, opened 2026-01-25
feat: integrate kicadts to generate centralized footprint library for build

Excellencedev:


MustafaMulla29 (commented):
I already told you that we are no longer using kicad-footprint-library. Don't just add it again. generateKicadFootprintLibrary is no longer needed.

Issue #1701, opened 2026-01-21
tsci build should use tscircuit.config.json settings


seveibar:

seveibar:
currently ignores stuff like kicad & kicadPcm

PR #1770, opened 2026-01-25
feat(build)!: consider tscircuit.config.json

ArnavK-09:
ref #1701


seveibar (approved):
Hard to maintain. There should be a function called resolveBuildOptions(…) that is used everywhere where we resolve build options (might already exist)

PR #1771, opened 2026-01-26
Cache the parts-engine in filesystem

imrishabh18:
image


lib/shared/get-complete-platform-config.ts
```diff
@@ -1,8 +1,35 @@
...
 import fs from "node:fs"
 
+export function createFilesystemCache(
+  cacheDir = path.join(process.cwd(), ".tscircuit", "cache", "parts-engine"),
+) {
+  return {
+    get: (key: string): string | null => {
+      try {
+        const hash = createHash("sha256").update(key).digest("hex")
+        const filePath = path.join(cacheDir, `${hash}.json`)
```
seveibar: make this semantically readable, it will save a lot of trouble long term

```suggestion
        const filePath = path.join(cacheDir, `${keyWithSafeCharacters.slice(0,20)}-${hash}.json`)
```

seveibar: also use md5 to hash, sha256 is not the right choice for file checksums

lib/shared/get-complete-platform-config.ts
```diff
@@ -1,8 +1,35 @@
 import type { PlatformConfig } from "@tscircuit/props"
 import { getPlatformConfig } from "@tscircuit/eval/platform-config"
+import { createHash } from "node:crypto"
 import path from "node:path"
 import fs from "node:fs"
 
+export function createFilesystemCache(
+  cacheDir = path.join(process.cwd(), ".tscircuit", "cache", "parts-engine"),
```
seveibar: this cache isn't just used by parts engine since it's labeled filesystemCache in the tscircuit platform config

```suggestion
  cacheDir = path.join(process.cwd(), ".tscircuit", "cache"),
```

lib/shared/get-complete-platform-config.ts
```diff
@@ -13,7 +40,9 @@ import fs from "node:fs"
 export function getCompletePlatformConfig(
   userConfig?: PlatformConfig,
 ): PlatformConfig {
-  const basePlatformConfig = getPlatformConfig()
+  const basePlatformConfig = getPlatformConfig({
+    filesystemCache: createFilesystemCache(),
+  })
```
seveibar: obfuscation!!!!!!!!

Look at this code, you have changed how it reads significantly.

The basePlatformConfig is the base platform config- it should not have this inside it!!!

seveibar (commented):
see comments

seveibar:
excited to use this!!!

lib/shared/get-complete-platform-config.ts
```diff
@@ -15,8 +50,13 @@ export function getCompletePlatformConfig(
 ): PlatformConfig {
   const basePlatformConfig = getPlatformConfig()
 
+  const platformConfigWithFilesystemCache = getPlatformConfig({
+    filesystemCache: createFilesystemCache(),
+  })
+
   const defaultConfig: PlatformConfig = {
     ...basePlatformConfig,
+    ...platformConfigWithFilesystemCache,
```
seveibar:
```suggestion
   filesystemCache: createFilesystemCache(),
```

lib/shared/get-complete-platform-config.ts
```diff
@@ -15,8 +50,13 @@ export function getCompletePlatformConfig(
 ): PlatformConfig {
   const basePlatformConfig = getPlatformConfig()
 
+  const platformConfigWithFilesystemCache = getPlatformConfig({
+    filesystemCache: createFilesystemCache(),
+  })
```
seveibar: Remove

seveibar (approved):

lib/shared/get-complete-platform-config.ts
imrishabh18: this needs to change in eval then

lib/shared/get-complete-platform-config.ts
seveibar: @imrishabh18 as per slack
```suggestion
    ...localCacheEngine: createFilesystemCache(),
```

PR #1794, opened 2026-01-27
Suppress the `console.log` for the project files

imrishabh18:
Note: This approach is modifying the global console object.

@seveibar What do you think about this approach? Another approach was to strip out the `console.log` on transpilation. But that will need significant refactor, to use rollup on build and will also add to latency


seveibar (changes requested):
We shouldnt modify console, the cloud platform should handle standard output better

PR #1818, opened 2026-01-29
fix: CLI build output options override config instead of merging

MustafaMulla29:
**Problem**: When running tsci build --kicad-library with a config file that has kicadPcm: true, both kicad-library AND kicad-pcm were being generated because CLI options were merging with config options.

**Fix**: Modified `cli/build/resolve-build-options.ts` to detect when any build output option is explicitly passed via CLI `(--kicad-library, --kicad-pcm, --preview-images, --site, etc.)`. When any CLI build output option is passed, config options are ignored entirely - CLI takes full control.

**New behavior**:
  - tsci build (no flags) → uses config options
  - tsci build --kicad-library → only builds kicad-library, ignores config
  - tsci build --kicad-library --kicad-pcm → builds both (explicit CLI flags)

### DIFF
--- a/cli/build/resolve-build-options.ts
+++ b/cli/build/resolve-build-options.ts
@@ -17,29 +17,53 @@ export const resolveBuildOptions = ({
 
   const configAppliedOpts: string[] = []
 
-  if (!cliOptions?.kicad && configBuild?.kicadLibrary) {
-    configAppliedOpts.push("kicad")
-  }
-  if (!cliOptions?.kicadLibrary && configBuild?.kicadLibrary) {
-    configAppliedOpts.push("kicad-library")
-  }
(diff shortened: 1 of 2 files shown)


seveibar:
What other cli has this behavior? Why is this correct? I dont think it is

MustafaMulla29:
When user runs tsci build --kicad-library, it's building both kicad-library and kicad-pcm as well by looking at the config options. Shouldn't the cli only build kicad-library if the user specifically put --kicad-library flag? @seveibar

seveibar:
That's not how boolean cli flags work. One boolean flag should not implicitly disable many others. What is the motivation for this feature? A flag like --only-kicad-library or --no-kicad-pcm is a convention, but im not sure we need this feature. You could also just add --ignore-config which is probably preferable. All the alternatives im suggesting are conventional for CLIs

PR #1879, opened 2026-02-02
fatalErrors should fail the whole build with exit code 1

imrishabh18:
Failing to generate the circuit json should throw a fatal error in `tsci build --ci`

Even the smoke test was failing for long time but the CI showed as passing
image

### DIFF
--- a/cli/build/build-file.ts
+++ b/cli/build/build-file.ts
@@ -9,6 +9,8 @@ import { getCompletePlatformConfig } from "lib/shared/get-complete-platform-conf
 export type BuildFileOutcome = {
   ok: boolean
   circuitJson?: unknown[]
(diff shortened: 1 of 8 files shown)


cli/build/build-file.ts
```diff
@@ -9,6 +9,8 @@ import { getCompletePlatformConfig } from "lib/shared/get-complete-platform-conf
 export type BuildFileOutcome = {
   ok: boolean
   circuitJson?: unknown[]
+  /** Fatal error that should always cause exit code 1, even with --ignore-errors */
+  fatalError?: boolean
```
seveibar:
```suggestion
  fatalError?: { errorType: "..." }
```

seveibar (approved):

PR #1936, opened 2026-02-06
fix: glb files should also be returned as relative path and not absolute from bun static asset loader

imrishabh18:
image


seveibar (commented):
Glb is a binary asset

seveibar (approved):

PR #1941, opened 2026-02-07
Differentiate error and warning in an agentically interpretable way

### LINKED ISSUE #1940: The distinction between error and warning is unclear to an agent when building
Image

This is GLM-4.7

Heinrich-XIAO:
Fixes: #1940


rushabhcodes (approved):
LGTM

ArnavK-09 (commented):
Should be more expressive?

Heinrich-XIAO:
> Should be more expressive?

The skill repo already said that warnings may not need to be addressed, so this is mostly a clarification to that.

PR #2027, opened 2026-02-16
Add kicadProject option to the build section

MustafaMulla29:
@seveibar the flag is ```tsci build --kicad```, is it ok to take the entrypoint from ```kicadLibraryEntrypoint``` or we should add a new entrypoint option?


seveibar (approved):

seveibar:
Please add an option! This behavior isnt clear

PR #2050, opened 2026-02-18
feat: implement camera positioning for 3D snapshots and add related tests

rushabhcodes:
This pull request improves the way 3D PCB snapshots are rendered by introducing a deterministic, board-size-aware camera positioning function. This ensures consistent and optimal camera angles for different board sizes across all environments. Additionally, a new test verifies correct snapshot creation for larger boards.

**Camera positioning improvements:**

* Added a new utility function `getBestCameraPosition` in `lib/shared/get-best-camera-position.ts` to calculate optimal, deterministic camera positions based on the board or panel size and center, ensuring consistent 3D rendering results.
* Updated `snapshotProject` in `lib/shared/snapshot-project.ts` to use `getBestCameraPosition` instead of hardcoded camera values, resulting in improved and consistent 3D snapshot images for all board sizes. [[1]](diffhunk://#diff-2092324af8269cfb0e148140806e9cd2bdaa13d8e031df470ee0204bc17b26b9R13) [[2]](diffhunk://#diff-2092324af8269cfb0e148140806e9cd2bdaa13d8e031df470ee0204bc17b26b9L144-R151)

**Testing enhancements:**

* Added a new test in `tests/cli/snapshot/snapshot.test.ts` to verify that the snapshot command correctly creates a 3D snapshot for a larger board, improving test coverage for the new camera logic.

## Before
before

## After 
image


lib/shared/get-best-camera-position.ts
```diff
@@ -0,0 +1,60 @@
...
+    | PcbBoard
+    | undefined
+
+  const surface = panel || board
+
+  if (!surface) {
+    // Default fallback for circuits without explicit board or panel
+    return {
+      camPos: [30, 30, 25] as const,
+      lookAt: [0, 0, 0] as const,
```
imrishabh18: The default position for the fallback should be `[10, 10, 10]`

imrishabh18 (changes requested):

lib/shared/get-best-camera-position.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  // Calculate camera distance based on board size
+  const maxDimension = Math.max(width, height)
+
+  // Use completely deterministic integer values to ensure identical rendering
+  // across all environments (local, CI, different Node versions, etc.)
+  const baseDistance = Math.round(maxDimension * 0.8)
```
imrishabh18: Maybe these numbers like `0.8` should be assigned a variable

imrishabh18 (approved):
Wait for @seveibar's review

seveibar (commented):
i think more examples are needed- to me it seems like it may be zoomed out too far, and the correct calculation would perfectly place the board inside the image. This can be done with some complex math

It is an improvement, but we might as well go for the most reliable long-term solution imo- especially because this is a breaking change (snapshots will be updated downstream)

seveibar:
also this method should be exported from circuit-json-to-gltf, it should not be in the CLI!!!!

seveibar (approved):

PR #2072, opened 2026-02-19
fix: improve error message when parts are not found in JLCPCB catalog

techmannih:
Screenshot_2026-02-19_23-13-25
Screenshot_2026-02-19_23-21-01


seveibar (changes requested):
misleading error always says it searched X when it didn't. Don't be verbose, don't link off platform

seveibar (approved):

PR #2180, opened 2026-03-02
Fix tsci init -y to suppress version update prompt in non-interactive mode

MustafaMulla29:


seveibar (changes requested):
make `shouldBeInteractive` detect the flag, will fix more at once and be simpler

lib/shared/check-for-cli-update.ts
```diff
@@ -7,11 +7,15 @@ import semver from "semver"
...
 
 export const currentCliVersion = () =>
   program?.version() ?? semver.inc(pkgVersion, "patch") ?? pkgVersion
 
 export const getLatestVersion = async () => {
+  if (process.env.TSCI_FAKE_LATEST_VERSION) {
+    return process.env.TSCI_FAKE_LATEST_VERSION
+  }
   const { version: latestCliVersion } = await ky
     .get<{ version: string }>(
```
seveibar, line 17: why are you running network checks if you're not going to act on them?

seveibar (changes requested):
slowness for no reason!

lib/shared/check-for-cli-update.ts
MustafaMulla29: oh ok sorry forgot it

seveibar (approved):

seveibar (approved):

PR #2255, opened 2026-03-05
Using the client server pattern for the worker (For a source file we generate the circuit-json, glbs and preview images according to the config)

imrishabh18:


seveibar (approved):

seveibar:
Remove refactor from title, refactor cannot contain functional changes

PR #2277, opened 2026-03-06
feat: add support for `--concurrency` for the `tsci snapshot` command

imrishabh18:
Have done some refactoring also to follow the DRY principle and reuse the `worker-pool` logic

Some tests
image


cli/snapshot/snapshot.worker.ts
```diff
@@ -0,0 +1,57 @@
+import { parentPort } from "node:worker_threads"
```
seveibar, line 1: *.worker.ts

seveibar (approved):

PR #2315, opened 2026-03-09
feat: add allowedSources policy to restrict tsci import sources

MustafaMulla29:
Ref https://github.com/tscircuit/agent-benchmarking-2026-02/issues/5

Screenshot_2026-03-09-19-17-36-44_40deb401b9ffe8e1df2f1cc5ba480b12

Adds a project-level allowedSources field to `tscircuit.config.json` that controls which sources tsci import is allowed to pull components from.

**Behaviour**
- Omitting allowedSources keeps the current behaviour — both JLCPCB and the tscircuit registry are searched
- Setting `allowedSources: ["jlcpcb"]` restricts import to JLCPCB only; using `--tscircuit` fails immediately with a clear error pointing to the config
- Setting `allowedSources: ["tscircuit"]` restricts to the tscircuit registry only; `--jlcpcb` and `--lcsc` are both blocked

### DIFF
--- a/cli/import/register.ts
+++ b/cli/import/register.ts
@@ -2,6 +2,7 @@ import type { Command } from "commander"
 import kleur from "kleur"
 import { getQueryFromParts } from "cli/utils/get-query-from-parts"
 import { importComponentFromJlcpcb } from "lib/import/import-component-from-jlcpcb"
+import { loadProjectConfig } from "lib/project-config"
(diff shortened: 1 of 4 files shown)


seveibar:
I mentioned in discord i didnt like this option, ai analysis is not always a good idea. The reason this is bad is because we are not allowed to change the directory from whatnusers have

there are other more clever solutions such as forcing a source in non interactive mode

PR #2372, opened 2026-03-12
Add periodic heartbeat logs for thread worker pool

imrishabh18:
### Motivation
- Provide periodic visibility into worker pool health so consumers can monitor busy/idle worker counts and queue depth.
- Follow an industry convention of emitting heartbeat-style status lines at a regular interval to aid debugging and observability.

### Description
- Add an optional `heartbeatIntervalMs` option to `ThreadWorkerPool` with a default of `5000`ms and emit heartbeat logs when `onLog` is provided.
- Heartbeat messages are emitted as `[worker-pool] heartbeat: workers busy=<busy>/<total>, idle=<idle>, queued_jobs=<n>` and are produced on a repeating interval started during initialization.
- Ensure heartbeat lifecycle is cleaned up by stopping the interval in both `stop()` and `terminate()` to avoid lingering timers.
- Add a unit test `tests/shared/thread-worker-pool-timeout.test.ts` that verifies heartbeat log emission using a short interval.

### Testing
- Ran `bun test tests/shared/thread-worker-pool-timeout.test.ts` which executed 2 tests and both passed.
- The heartbeat test asserted that logs contain the `[worker-pool] heartbeat:` string and completed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69b30dfe0fcc8327a58d53806449dca9)

### DIFF
--- a/lib/shared/thread-worker-pool.ts
+++ b/lib/shared/thread-worker-pool.ts
@@ -25,6 +25,7 @@ type ThreadWorkerPoolOptions<TJob, TWorkerInput, TWorkerOutput, TResult> = {
   onLog?: (lines: string[]) => void
   cancellationError?: Error
   jobTimeoutMs?: number
+  heartbeatIntervalMs?: number
 }
 
 export class ThreadWorkerPool<TJob, TWorkerInput, TWorkerOutput, TResult> {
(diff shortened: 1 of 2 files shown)


seveibar (approved):
I would log the specific worker tasks for each worker and how long theyve been running for, its not bery useful for debugging as is

PR #2390, opened 2026-03-13
Enforce deterministic no-match failures for tsci snapshot <directory> and surface exact includeBoardFiles source

AnasSarkiz:
## Summary
`tsci snapshot <directory>` previously returned exit `0` with a generic `No entrypoint found. Run 'tsci init' to bootstrap a project or specify a file with 'tsci snapshot <file>` when the directory had no matching board/circuit files.

This change makes explicit directory no-match behavior strict and clear:
- exit code is now `1`
- error shows which `includeBoardFiles` patterns were used
- source is explicit: config vs default patterns
- no-target snapshot behavior stays unchanged

 Terminal Results (Before -> After)

 1) Directory has matching files
 tsci snapshot dir --update
 before: exit 0, Created snapshots
 after : exit 0, Created snapshots

 2) Directory no-match + config includeBoardFiles
 tsci snapshot src
 before: exit 0, `"No entrypoint found..."`
 after : exit 1
`No circuit files found to create snapshots in directory: "src"
Searched using tscircuit.config.json includeBoardFiles: ["src/**/*.board.tsx"]`

 3) Directory no-match + default patterns
$ tsci snapshot lib/example_connectors/example_header --concurrency 4
 before: `exit 0, "No entrypoint found..."`
 after : `exit 1`
`No circuit files found to create snapshots in directory: "lib/example_connectors/example_header"
Searched using default includeBoardFiles: ["**/*.board.tsx","**/*.circuit.tsx","**/*.circuit.json"]`

### DIFF
--- a/lib/shared/snapshot-project.ts
+++ b/lib/shared/snapshot-project.ts
@@ -42,13 +42,13 @@ type SnapshotOptions = {
   onSuccess?: (message: string) => void
 }
 
-const getIncludeBoardFilesSource = (projectDir: string): boolean => {
+const hasConfiguredIncludeBoardFiles = (projectDir: string): boolean => {
   const projectConfig = loadProjectConfig(projectDir)
-  const hasConfiguredIncludeBoardFiles = Boolean(
+  const hasConfiguredPatterns = Boolean(
     projectConfig?.includeBoardFiles?.some((pattern) => pattern.trim()),
   )
 
-  return hasConfiguredIncludeBoardFiles
+  return hasConfiguredPatterns
 }
 
 export const snapshotProject = async ({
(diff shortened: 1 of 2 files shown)


lib/shared/snapshot-project.ts
```diff
@@ -37,6 +42,15 @@ type SnapshotOptions = {
   onSuccess?: (message: string) => void
 }
 
+const getIncludeBoardFilesSource = (projectDir: string): boolean => {
```
seveibar: Confusing function name? 

Implies it returns inckudedBoardFiles

seveibar (approved):
Rename function, get someone to check it !

ShiboSoftwareDev (approved):

PR #2471, opened 2026-03-19
feat: add support --download flag to downloading and localizing 3D models from JLCPCB

techmannih:
This PR adds the ability to download and localize 3D models (OBJ files) when importing components from JLCPCB using the `tsci import` command.

## Summary of Changes
- **CLI Enhancement**: Added a new `--download` flag to the `tsci import` command.
- **3D Model Localization**: Implemented logic to fetch `.obj` files directly from JLCPCB's CDN.
- **Component Linking**: Automatically updates the generated TypeScript component to use the relative path of the downloaded 3D model instead of a remote URL.
- **Filename Resolution**: Includes a robust parser for the `content-disposition` header to ensure the downloaded file uses the correct original filename from JLCPCB.
Screenshot_2026-03-19_10-46-52


lib/import/import-component-from-jlcpcb.ts
```diff
@@ -2,19 +2,103 @@ import { fetchEasyEDAComponent, convertRawEasyToTsx } from "easyeda/browser"
...
+  const remoteUrlMatch = objUrlMatch || modelUrlMatch
+
+  if (!remoteUrlMatch) {
+    return tsxContent
+  }
+
+  const remoteUrl = remoteUrlMatch[1]
+
+  try {
+    const response = await fetch(remoteUrl)
```
MustafaMulla29: ig you should be using `platformFetch` instead of fetch

cli/import/register.ts
```diff
@@ -144,13 +146,14 @@ export const registerImport = (program: Command) => {
             return process.exit(1)
           }
         } else {
+          const partId = `C${choice.part}`
```
seveibar: Unclear naming here, rename part and partId

partNumber is fine. lcscId is better than partId and used elsewhere

seveibar (changes requested):
Regex on tsx is a hack

seveibar (commented):
always avoid using regex unless it greatly simplifies a solution, regex should never make a solution more complex (it makes it more complex here)

The easyeda converter converts circuit json into tsx internally, can't you use that to simplify this implementation and avoid regex.

Can you also make sure to properly handle the step model url, which was in the original feature request Ray made (this wouldn't have helped him because you didn't handle step model urls)

Why do you handle multiple remote urls in this way? Simplify your code, use .find instead of filter to find the cad component.

PR #2476, opened 2026-03-19
Integrate `@tscircuit/image-utils` and remove sharp and looks-same from the dependency

imrishabh18:


seveibar (approved):

seveibar:
Is this tested properly?

imrishabh18:
> Is this tested properly?

Not sure about properly, but running against the cli which has a lot of tests for the `tsci snapshot` which uses this method. I think this should be good?

PR #2508, opened 2026-03-23
fix: upgrade prompt never showing when tsci is outdated

MustafaMulla29:
`currentCliVersion()` imports program from cli/main but `.version()` is never set on the commander program (version flags are handled manually). This caused `program.version()` to always return undefined, falling back to `semver.inc(pkgVersion, "patch")` — one patch above the real version. Since the reported version was always artificially higher `semver.gt(latestNpmVersion, currentCliVersion())`was always false and the update prompt never fired. Fixed by falling back directly to pkgVersion.

### DIFF
--- a/lib/shared/check-for-cli-update.ts
+++ b/lib/shared/check-for-cli-update.ts
@@ -9,8 +9,7 @@ import kleur from "kleur"
 import { prompts } from "lib/utils/prompts"
 import { shouldBeInteractive } from "lib/utils/should-be-interactive"
 
-export const currentCliVersion = () =>
-  program?.version() ?? semver.inc(pkgVersion, "patch") ?? pkgVersion
+export const currentCliVersion = () => program?.version() ?? pkgVersion
 
 export const getLatestVersion = async () => {
   const { version: latestCliVersion } = await ky
--- a/tests/cli/upgrade/check-for-cli-update.test.ts
+++ b/tests/cli/upgrade/check-for-cli-update.test.ts
@@ -1,0 +1,22 @@
+import { expect, test } from "bun:test"
+import semver from "semver"
+import { currentCliVersion } from "lib/shared/check-for-cli-update"
+import { version as pkgVersion } from "../../../package.json"
+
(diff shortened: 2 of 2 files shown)


seveibar (commented):
I don’t think this is the right fix. You have to keep in mind that tscircuit version is much different than the CLI version.

if this solution is correct, can you make the code more explicit? Can you actually distinguish between tscircuit version and the CLI version

MustafaMulla29:
Screenshot 2026-03-24 at 12 18 12 AM

@seveibar i think its working?

seveibar:
My suggestion is mostly to clarify variable names and maybe avoid all the ??, there is ONE SOURCE OF TRUTH for the version, use that and name variables appropriately to reflect the actual package being version compared

seveibar:
I doubt this will work once published!

MustafaMulla29:
> I doubt this will work once published!

No, I said it's already working. I think this is  also not the correct fix. It's working on bunx but when we explicitly install outdated version and try without bunx it's not working @seveibar

seveibar:
@MustafaMulla29 FWIW i have no idea what you're talking about because of the lack of specificity.

there is the possible installation of either @tscircuit/cli or "tscircuit"

The symlink used for bunx could different from the bin from tscircuit, it sounds like you're on it but try to be specific otherwise i'm going to have to write a long comment like this and think too hard about what you could mean

MustafaMulla29:
> @MustafaMulla29 FWIW i have no idea what you're talking about because of the lack of specificity.
> 
> there is the possible installation of either @tscircuit/cli or "tscircuit"
> 
> The symlink used for bunx could different from the bin from tscircuit, it sounds like you're on it but try to be specific otherwise i'm going to have to write a long comment like this and think too hard about what you could mean

I think we have to install cli as a seperate package to make tsci work and also install tscircuit seperately. What I mean is I tried with an outdated tscircuit dep in a tsci project with bunx and it prompts there to update(as you can see in the image). But when I do it without bunx and with an outdated tscircuit and cli package, it doesn't prompt 

@seveibar

PR #2509, opened 2026-03-23
Feat: download flag for downloading and localizing 3d models from JLCPCB

MustafaMulla29:


seveibar (approved):
Can yiu make sure were properly copying iver to kicad

PR #2569, opened 2026-03-29
Fix: Show Cli Version Instead of tscircuit Version

mohan-bee:
this is confusing for the user which cli version is installed this helps tracking based on cli version but not on tscircuit version.

image

image


mohan-bee:
can i also modify the test according to this !

seveibar (changes requested):
CC @MustafaMulla29 

End users ONLY track and install the version of tscircuit

PR #2570, opened 2026-03-29
Log `asyncEffect:start` events in `tsci build` process

imrishabh18:


lib/shared/generate-circuit-json.tsx
```diff
@@ -30,6 +30,18 @@ type GenerateCircuitJsonOptions = {
...
 
+const attachAsyncStatusLogger = (circuit: {
+  on: (
+    event: "asyncEffect:start",
+    listener: (event: { effectName?: string }) => void,
+  ) => void
+}) => {
+  circuit.on("asyncEffect:start", ({ effectName }) => {
+    const asyncEffectName = effectName || "unknown async effect"
+    console.log(`waiting on ${asyncEffectName}…`)
```
seveibar, line 41: so you need to track the start/end or track the active async effects, this isn't a good way to log it because you're not actually waiting on the async effect

PR #2577, opened 2026-03-31
Fix gerber export case missing

mohan-bee:
Previously this gerber export is not defined in the switch case so it is defaults to circuit json as gerber is .zip it wraps json into .zip 

https://github.com/user-attachments/assets/b9d56591-ce1c-41bb-9d6b-5c20495d0ff3

So this PR adds a new switch case to handle gerber export properly. which contains drill export for plated holes , drill export for non plated holes, included with BOM.csv and pick_and_place.csv.

https://github.com/user-attachments/assets/8f916595-51b1-4193-a212-7b6b7a5b8a1b

https://github.com/user-attachments/assets/9057ebc2-be27-424e-b065-8b7e5810a087


seveibar (approved):

seveibar:
Thanks @mohan-bee !!!!!

PR #2592, opened 2026-04-02
Fix build ensures proper exits if any unignored build errors

mohan-bee:
This PR Ensures `tsci build` fails and exits with code 1 when build errors occur. 

Fixed build to return non-zero exit code for unignored routing and DRC errors


MustafaMulla29 (commented):
can you create one more test that builds with mutliple errors. Inlinesnapshot the full build output.

cli/build/register.ts
```diff
@@ -952,7 +959,12 @@ export const registerBuild = (program: Command) => {
             : kleur.green("\n✓ Done"),
         )
         if (shouldExitNonZero) {
-          exitBuild(1, "fatal circuit build errors occurred")
+          exitBuild(
+            1,
+            hasFatalErrors
+              ? "fatal circuit build errors occurred"
+              : "circuit build errors occurred",
```
MustafaMulla29: This can be "build finished with errors" instead of "circuit build errors occurred"

tests/cli/build/build-with-multiple-errors.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    Port pin1 on R1 is missing a trace
+    Port pin2 on R1 is missing a trace
+    Port pin2 on R1 is missing a trace
+    Port pin1 on R2 is missing a trace
+    Port pin1 on R2 is missing a trace
+    Port pin2 on R2 is missing a trace
+    Port pin2 on R2 is missing a trace
+
+    Build complete
+      Circuits  0 passed 1 with errors
```
MustafaMulla29, line 42: I think since the artifacts are produced either way, we should not say "0 passed". Technically the assets are generated. So ig it should be eg: "1 passed with errors"(for only one circuit), "3 passed 1 with errors". I might be wrong here, @seveibar need your verdict.

MustafaMulla29 (changes requested):
See comment

tests/cli/build/build-with-multiple-errors.test.ts
mohan-bee: The word "passed" in a build context specifically means passed all checks. Showing 1 passed with errors creates a contradiction in terms and could mislead a user to  manufatue a broken board. 0 passed 1 with errors is like "nothing is fabrication ready".

mohan-bee: image


as you can see the passed and errors are always separate counts, never combined.

tests/cli/build/build-ignore-drc-categories.test.ts
```diff
@@ -47,19 +47,20 @@ test("build reports ignored counts when selected DRC categories are filtered", a
   )
   await writeFile(path.join(tmpDir, "package.json"), "{}")
 
-  const { exitCode, stdout } = await runCommand(
+  const { exitCode, stdout, stderr } = await runCommand(
     "tsci build board.circuit.json --ignore-placement-drc --ignore-routing-drc",
```
seveibar, line 52: I dont understand- isnt the user requesting that certain errors be ignored?

tests/cli/build/build-with-multiple-errors.test.ts
seveibar: Many users have requested to be able to ignore build errors, this is provided via flags, it seems like you’re making the ignore flags not work?

seveibar (commented):
The motivation of the author isnt clear. Changing tests without considering the intent?

mohan-bee:
The --ignore-*-drc flags still work correctly (ignored errors are filtered out and don't affect exit code)  

the first test exits 1 only because of the remaining unignored netlist error

I've added comments clarifying this and new tests proving both --ignore-errors and --ignore-*-drc flags work as expected.

cli/build/register.ts
```diff
@@ -876,11 +877,16 @@ export const registerBuild = (program: Command) => {
           }
         }
 
-        // Fatal errors (e.g., circuit generation exceptions) always cause exit code 1.
-        const shouldExitNonZero = hasFatalErrors
+        // Exit non-zero on fatal errors or unignored DRC errors (--ignore-errors / --ignore-*-drc flags suppress accordingly)
+        const shouldExitNonZero = hasFatalErrors || hasErrors
```
seveibar, line 881: Confusing comment, the code beneath it doesn't make sense

seveibar:
@mohan-bee you're confusing me.

Create a reproduction WITHOUT IMPLEMENTATION, the reproduction should demonstrate the issue you are fixing, the test should assert the incorrect behavior with a comment. After the reproduction is merged, you can merge a fix

MustafaMulla29:
I think he meant to fix the exit code issue, It's returning exit code 0 with "build finished successfully" even when there are errors(such as auto routing, drc etc)

PR #2622, opened 2026-04-09
Fix wrong shell chaining command

mohan-bee:


seveibar (approved):

seveibar:
best practice is to add a test so nobody breaks this functionality in the future- the original author should have done it

mohan-bee:
> best practice is to add a test so nobody breaks this functionality in the future- the original author should have done it

ok let me add that !

PR #2885, opened 2026-04-30
test: add tests for SVG output location (issue #2550)

64johnlee:
## Issue
Issue #2550: SVGs placed at dist root instead of component subdirectory

## Tests Added
These tests verify that:
1. SVGs are placed alongside circuit.json in the component subdirectory
2. Multiple file builds with --all-images work correctly
3. Files in same-named directories work

## Test Results  
Tests pass in single-threaded mode. Parallel mode (concurrency > 1) has environment dependency issues.

/claim #2550

### DIFF
--- a/cli/check/routing/register.ts
+++ b/cli/check/routing/register.ts
@@ -1,11 +1,104 @@
+import {
+  categorizeErrorOrWarning,
+  type DrcCategory,
+} from "@tscircuit/circuit-json-util"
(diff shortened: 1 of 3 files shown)


rushabhcodes:
link the correct issue, also  format check failing

rushabhcodes:
link the correct issue, also  format check failing

PR #2913, opened 2026-05-02
fix: add fallback to find any .tsx/.ts/.circuit.json when no standard entrypoint found

### LINKED ISSUE #2797: `tsci push` fails when no `index.circuit.tsx` file and no `mainEntrypoint`

64johnlee:
## Summary

`tsci push` fails when no `index.circuit.tsx` (or other standard entrypoint) exists AND no `mainEntrypoint` is configured in `tscircuit.config.json`. However, `tsci dev` finds alternative files — so push should do the same.

## Root Cause

`push-snippet.ts` calls `getEntrypoint()` and immediately exits if it returns `null`. Meanwhile, `resolve-dev-target.ts` has a fallback that uses `findSelectableFiles()` to locate any `.tsx/.ts/.circuit.json` file when `getEntrypoint()` fails.

## Fix

Added the same fallback logic to `pushSnippet()`:

```typescript
if (!snippetFilePath) {
  const fallbackFiles = globbySync(
    ["**/*.tsx", "**/*.ts", "**/*.circuit.json"],
    {
      cwd: process.cwd(),
      ignore: DEFAULT_IGNORED_PATTERNS,
    },
  )
  if (fallbackFiles.length > 0) {
    const fallbackPath = path.resolve(process.cwd(), fallbackFiles[0])
    console.log(kleur.gray(`No standard entrypoint found. Using fallback file: ...`))
    snippetFilePath = fallbackPath
  }
}
```

This keeps the priority: explicit `filePath` argument > `mainEntrypoint` config > standard entrypoint names > **fallback to any circuit file**.

## Testing
- Existing tests pass
- Logic mirrors `tsci dev` fallback behavior

## Related
- Fixes #2797
- cf. `resolve-dev-target.ts` which already has this fallback

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,3 +84,4 @@ The CLI entrypoint (`cli/entrypoint.js`) selects between Bun and tsx as the Type
 # bump 1777865732
 # bump 1777867214
 # bump 1777910427
+# bump 1777953613
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,3 +81,4 @@ Test fixture provides:
 
 The CLI entrypoint (`cli/entrypoint.js`) selects between Bun and tsx as the TypeScript runner, preferring Bun when available. This allows hot-reload during development while maintaining Node.js compatibility.
 # bump 1777865644
+# bump 1777865732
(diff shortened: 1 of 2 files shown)


64johnlee:
Duplicate of #2977, closing

PR #2914, opened 2026-05-02
fix: support board files as entrypoint for tsci push

### LINKED ISSUE #2797: `tsci push` fails when no `index.circuit.tsx` file and no `mainEntrypoint`

64johnlee:
Like tsci dev, tsci push should find board files (*.circuit.tsx, *.board.tsx, *.circuit.json) when no standard entrypoint exists.

Fixes #2797

## Changes
- Added findBoardFiles import to push-snippet.ts
- After failing to find a standard entrypoint, fall back to finding board files (same logic as tsci dev)
- Only show the No entrypoint found error if both searches fail

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,3 +87,4 @@ The CLI entrypoint (`cli/entrypoint.js`) selects between Bun and tsx as the Type
 # bump 1777953610
 # bump 1777996811
 # bump 1778040016
+# bump 1778083219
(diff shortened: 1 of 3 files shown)


64johnlee (commented):
All CI checks are passing! This PR adds board file fallback support for tsci push, matching tsci dev behavior. Ready for merge. 🙏

PR #2915, opened 2026-05-02
 Make tsci push fall back to selectable project files when no entrypoint is configured

### LINKED ISSUE #2797: `tsci push` fails when no `index.circuit.tsx` file and no `mainEntrypoint`

rushabhcodes:
tsci push previously failed with a "No entrypoint found" error when a project had no index.circuit.tsx and no mainEntrypoint, even though tsci dev already handled that case by falling back to the first
  selectable project file.

  This PR aligns push with dev by reusing the same fallback behavior:

  - try the configured or detected entrypoint first
  - if none exists, fall back to board files
  - then fall back to selectable .tsx, .ts, or .circuit.json files

  It also adds a regression test covering a project with no conventional entrypoint, verifying that tsci push succeeds and does not print the entrypoint error.

  Why this should score well against the guide:

  - it is a real bug fix, not a vague cleanup
  - it includes a reproduction test and the fix in the same PR
  - it is a meaningful CLI UX improvement
  
  /fixes #2797

### DIFF
--- a/lib/shared/push-snippet.ts
+++ b/lib/shared/push-snippet.ts
@@ -15,6 +15,9 @@ import { checkOrgAccess } from "lib/utils/check-org-access"
 import { isBinaryFile } from "./is-binary-file"
 import { hasBinaryContent } from "./has-binary-content"
 import JSZip from "jszip"
+import { findBoardFiles } from "./find-board-files"
+import { globbySync } from "globby"
+import { DEFAULT_IGNORED_PATTERNS } from "./should-ignore-path"
 
 type PushOptions = {
   filePath?: string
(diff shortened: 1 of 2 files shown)


lib/shared/push-snippet.ts
```diff
@@ -30,6 +33,26 @@ type PushOptions = {
...
+const findSelectableFiles = (projectDir: string): string[] => {
+  const boardFiles = findBoardFiles({ projectDir })
+    .filter((file) => fs.existsSync(file))
+    .sort()
+
+  if (boardFiles.length > 0) {
+    return boardFiles
+  }
+
+  const files = globbySync(["**/*.tsx", "**/*.ts", "**/*.circuit.json"], {
```
seveibar, line 45: you need to use the project config for this i think

seveibar (commented):
selecting a random file is kind of silly, what logic does runframe use to pick the default entrypoint?

rushabhcodes:
i copied this logic from `tsci dev` cause @imrishabh18 said it works for it.

PR #2924, opened 2026-05-04
fix: allow tsci push without entrypoint when circuit.json exists

64johnlee:
## Summary

Allow `tsci push` to work without an explicit entrypoint when a `circuit.json` file exists in the project root.

## Changes

- Fallback to circuit.json when no entrypoint is specified
- Works with projects that have circuit.json but no entrypoint configured


64johnlee:
Duplicate of #2952, closing

PR #3080, opened 2026-05-19
Add telemetry for the `tsci init` command

imrishabh18:
image


seveibar (changes requested):
This pattern is too complex, why the with*? Just allow people to emit events without changing wrapping code

package.json
```diff
@@ -77,6 +77,7 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
+    "circuit-json": "^0.0.425",
```
seveibar: ?

seveibar (commented):
Error tracking and usage telemetry are separate concerns

seveibar:
I can understand wanting to track failures and it makes sense, but i think we're mixing concerns and creating a complex pattern- i would prefer, for example, that we do what bun does and generate report urls on any cli failure rather than instrument failure catchers everywhere where theyre often unwanted

package.json
imrishabh18: smoke tests have been failing cause of some old version of circuit-json in peer deps is being used.

This was a fix for that, while on the side I have been trying to update circuit-json version in all the down stream packages

imrishabh18: The upstream pipeline is blocked because of this

imrishabh18: Fixed now, with updating the tscircuit to the latest version

seveibar (approved):

PR #3097, opened 2026-05-23
Send telemetry info using specific id's associated to users to calculate DAU/WAU/MAU in posthog

imrishabh18:


seveibar (commented):
This will make the cli slower (a lot slower), it needs to be done carefully so that telemetry does not make commands run more slowly. also it's capturing all the ci usage i think- we can probably do some basic checks so that ci usage doesnt get sent rather than filter on the posthog side

Do we have events in runframe yet? Runframe and circuit json viewer telemetry may be a bit better

lib/telemetry/index.ts
```diff
@@ -71,6 +86,7 @@ export const captureTelemetryEvent = async (
         "Content-Type": "application/json",
       },
       body: JSON.stringify(payload),
+      signal: AbortSignal.timeout(1_000),
```
seveibar, line 89: Each cli command taking 1s is wayyyy too long. An abortsignal needs to be sent at the end of every command

PR #3149, opened 2026-06-01
Add runtime tscircuit config loader support

techmannih:
## Summary

This PR adds runtime config module support to the CLI so projects can define configuration in:

- `tscircuit.config.ts`
- `tscircuit.config.js`

The loader now also:

- keeps existing `tscircuit.config.json` support
- merges JSON config with module config when both exist
- loads `.env` and `.env.local` before importing the config module
- preserves `platformConfig` as a runtime-only field outside the JSON schema

This is the first, narrow step toward supporting custom parts/platform behavior through project configuration.

## Why

The current CLI config flow is centered around `tscircuit.config.json`, which only supports serializable JSON values.

For custom runtime behavior, that is not enough. A project may need to provide config that depends on:

- functions
- imported helpers
- environment variables such as `PARTNER_TOKEN`

### DIFF
--- a/lib/project-config/index.ts
+++ b/lib/project-config/index.ts
@@ -1,5 +1,7 @@
 import * as fs from "node:fs"
 import * as path from "node:path"
+import { pathToFileURL } from "node:url"
+import type { PlatformConfig } from "@tscircuit/props"
 import {
   projectConfigSchema,
   type TscircuitProjectConfig,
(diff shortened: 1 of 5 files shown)


seveibar (approved):
CC @imrishabh18

lib/project-config/index.ts
```diff
@@ -21,10 +32,74 @@ export const DEFAULT_BOARD_FILE_PATTERNS = [
   "**/*.circuit.json",
 ]
 
+const parseProjectConfigObject = (
+  config: unknown,
```
imrishabh18, line 36: Why `unknown`?

techmannih: Because it is runtime config input, so I treated it as untrusted and narrowed it before reading fields.

imrishabh18 (approved):

PR #3291, opened 2026-06-12
Add ti flag for tsci search command

MustafaMulla29:


tests/cli/search/search-ti.test.ts
imrishabh18, line 1: Use the fake for testing, checkout the `ti-parts-engine` on how the fake is being used to test

imrishabh18 (changes requested):

tests/cli/search/search-ti.test.ts
MustafaMulla29: Is using `fake-ul-kicad-proxy` correct in cli?

cli/search/register.ts
```diff
@@ -194,6 +221,31 @@ export const registerSearch = (program: Command) => {
             )
           })
         }
+
+        if (tiResults.length) {
+          console.log()
```
imrishabh18:
```suggestion
```

imrishabh18 (approved):

PR #3362, opened 2026-06-18
 Add simulation SVG outputs for build and snapshots

ShiboSoftwareDev:
Adds build flags for simulation.svg and schematic-simulation.svg, plus snapshot support for
  *-simulation.snap.svg and *-schematic-simulation.snap.svg. Simulation SVGs are generated only when
  circuit JSON already contains analog simulation graph results, so normal circuits do not run or
  emit simulation outputs. Updates simulation-related dependencies, switches snapshot diffing to
  direct looks-same, and adds TSX boost converter coverage for build and snapshot generation.


package.json
```diff
@@ -12,15 +12,15 @@
     "@tscircuit/circuit-json-placement-analysis": "^0.0.6",
     "@tscircuit/circuit-json-routing-analysis": "^0.0.1",
     "@tscircuit/circuit-json-schematic-placement-analysis": "github:tscircuit/circuit-json-schematic-placement-analysis#b6d5fe675adc26a2fc988aa081e6fb51e6bf696a",
-    "@tscircuit/eval": "^0.0.915",
+    "@tscircuit/eecircuit-engine": "https://jscdn.tscircuit.com/@tscircuit/eecircuit-engine/1.7.4.tgz",
```
imrishabh18, line 15: Why not add this package in the `tscircuit` and don't have to add it on all the other package?

ShiboSoftwareDev: @imrishabh18 Seve said I can't add it to tscircuit, I can try to find another way to resolve this later

cli/build/register.ts
```diff
@@ -149,6 +149,14 @@ export const registerBuild = (program: Command) => {
       "--schematic-svgs",
       "Generate schematic SVG outputs during build generation",
     )
+    .option(
+      "--simulation-svgs",
+      "Generate simulation graph SVG outputs during build generation",
+    )
+    .option(
+      "--schematic-simulation-svgs",
+      "Generate schematic simulation SVG outputs during build generation",
```
imrishabh18, line 158: The flag's are confusing, what is the difference between them?

ShiboSoftwareDev: @imrishabh18 one of them is a simulation graph, and one of them is a graph with the schematic

imrishabh18 (approved):

PR #3363, opened 2026-06-19
feat: add layer specific snapshot for top and bottom

mohan-bee:
**Motivation**: snapshot PCB layers individually for clearer visual testing.

**Adds:**
`tsci snapshot --layer top --pcb-only` or `tsci snapshot --layer top `
`tsci snapshot --layer bottom --pcb-only` or `tsci snapshot --layer bottom `

`--layer <top | bottom>`implies for PCB only behaviour. 

This creates separate top or bottom PCB SVG snapshots without changing existing snapshot behavior.

hopefully this won't break anything also for now this only supports for 2 layers(e.g top and bottom).


imrishabh18 (approved):
It should also support other layers when the board has more than 2 layer.

Get a review from @seveibar before merging

mohan-bee:
> It should also support other layers when the board has more than 2 layer.
> 
> Get a review from @seveibar before merging

yep, i was starting with 2 layer.  will add more ..

seveibar (approved):

PR #3378, opened 2026-06-19
Support adding multiple packages with add command

r-bt:
While following the tutorial at: https://docs.tscircuit.com/tutorials/build-a-custom-keyboard-with-tscircuit, it says to run:
```bash
tsci add @tsci/seveibar.PICO @tsci/seveibar.Key
```

However this throws an error:
```bash
error: too many arguments for 'add'. Expected 1 argument but got 2.
```

Using the variadic argument `<packageSpecs...>` in the commander definition resolves this. 

Note that adding multiple packages through the interactive cli (i.e. running `tsci` and then selecting add from the menu) did work.

However, this highlighted a bug with the `addPackage` function where it was only normalizing the first package in a list of packages. So I added a new function `addPackages` which normalizes all the packages, and have `addPackage` just call that.

All the existing add testcases pass, and I added a new one to verify that adding two packages works

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Commands:
   build [options] [file]       Run tscircuit eval and output circuit json
   transpile [file]             Transpile TypeScript/TSX to JavaScript (ESM,
                                CommonJS, and type declarations)
-  add <packageSpec>            Add a tscircuit component package to your project
+  add <packageSpecs...>        Add tscircuit component packages to your project
   remove <component>           Remove a tscircuit component package from your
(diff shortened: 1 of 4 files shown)


imrishabh18 (approved):

mohan-bee (approved):

rushabhcodes (commented):
also update in readme

r-bt:
> also update in readme

Done

rushabhcodes (approved):

PR #3457, opened 2026-06-25
Align simulate SPICE platform config with snapshot behavior

ShiboSoftwareDev:


imrishabh18 (commented):
I don't think the naming convention is right for the function, what does `complete` mean here?

imrishabh18 (approved):

PR #3469, opened 2026-06-25
Fix: tsci dev uses the project-local tscircuit version when available

MustafaMulla29:
`tsci dev` now renders the preview with the `tscircuit` version installed in the project, instead of always using the runframe baked into the CLI.

Previously the dev server served its own bundled `@tscircuit/runframe/standalone` (pinned to the CLI's devDependency, with eval fetched from the CDN at runtime) for every project — so a project pinning a specific `tscircuit` version was ignored.

## How

The `/standalone.min.js` route in `lib/server/createHttpServer.ts` now resolves `tscircuit` from the project dir (`createRequire(projectDir/package.json)`) and, if found, serves that package's `dist/browser.min.js` — the self-contained runframe + eval bundle with the matching `@tscircuit/core`/`@tscircuit/eval`  baked in.

Resolution order:
1. `RUNFRAME_STANDALONE_FILE_PATH` env override (unchanged)
2. project-local `tscircuit/dist/browser.min.js` (new)
3. the runframe bundled into the CLI (existing fallback)

No changes needed in the `tscircuit` or `runframe` repos — `tscircuit` already ships the bundle; the CLI just wasn't looking for it.

### DIFF
--- a/lib/server/createHttpServer.ts
+++ b/lib/server/createHttpServer.ts
@@ -1,5 +1,7 @@
 import * as fs from "node:fs"
 import * as http from "node:http"
+import { createRequire } from "node:module"
+import * as path from "node:path"
 // @ts-ignore
 import runFrameStandaloneBundleContent from "@tscircuit/runframe/standalone" with {
   type: "text",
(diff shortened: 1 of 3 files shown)


seveibar (approved):

seveibar:
@MustafaMulla29 note: a potentially better fix (complementary , both can be merged) is in tscircuit/tscircuit by adjusting the cli binary that is called when "tsci" is used

PR #3485, opened 2026-06-29
Add Gerber-based PCB short circuit detection

Abse2001:
Motivation
Shorts can be introduced or missed if we only inspect pre-Gerber circuit JSON geometry. This check validates the copper geometry that is actually emitted for fabrication, including traces, pads, vias, plated holes, and copper pours.
Implementation
Added tsci check short [file].
Generates Gerber ZIP output for the checked board.
Reconstructs copper primitives from Gerber commands:aperture flashes
stroked traces
filled regions
copper-pour clear regions/cutouts

Detects same-layer intersections between copper assigned to different nets.
Adds Gerber X2 net metadata using %TO.N,<net>*% / %TD.N*%.
Uses X2 net metadata during analysis when present.
Falls back to circuit-json geometry matching for object/net association.
Splits short-check registration, analysis, and geometry helpers into dedicated files following CLI command conventions.


seveibar (changes requested):
Needs rewrite

* reimplementing a ton of geometry checks and gerber types
* implementation for this lives in cli without robust testing
* No bitmap based testing
* Any new shapes etc require new implementation

Do this:
Create repo called check-shorts
Implement robust repros
Use circuit-to-canvas and do “bitmap coloring” using pcb_port and connectivity map keys. If you ever have to recolor with a connectivity map key then you have a short
Include a gerber mode which internally creates RASTERIZED gerbers

PR #3834, opened 2026-07-24
Add chip-focused PCB snapshots

seveibar:
## What

- add `tsci snapshot --chip <name>` to crop PCB snapshots around a named source component
- make `--chip` imply `--pcb-only` and retain nearby traces, vias, pads, and board geometry inside the viewport
- add `--chip-padding <millimeters>` with a 2 mm default
- support focused top/bottom layer snapshots through the existing `--layer` option
- write focused snapshots with distinct names such as `board-pcb-U1.snap.svg`
- report available PCB component names when the requested component cannot be found

## Why

Dense packages such as BGAs are difficult to inspect in whole-board snapshots. A component-focused viewport makes escape routing, clearance issues, and apparent layer crossings legible without post-processing the SVG.

## Impact

This is additive and does not change existing snapshot output unless `--chip` is supplied. The option is intentionally PCB-only; unsupported schematic, simulation, and 3D combinations fail with a clear message.

## Checks

- `bun test tests/shared/get-pcb-component-viewport.test.ts tests/cli/snapshot/snapshot-chip.test.ts`
- `bun run format:check`
- `bunx tsc --noEmit`
- `bun run build`
- manually generated all-layer, top-only, and bottom-only focused snapshots for a 144-pin BGA autorouting result

### DIFF
--- a/cli/snapshot/register.ts
+++ b/cli/snapshot/register.ts
@@ -23,6 +23,14 @@ export const registerSnapshot = (program: Command) => {
       "--layer <layer>",
       "Generate a PCB snapshot for one layer: top or bottom (implies --pcb-only)",
     )
+    .option(
+      "--chip <name>",
+      "Crop the PCB snapshot around a named chip/component (implies --pcb-only)",
+    )
+    .option(
+      "--chip-padding <millimeters>",
+      "Padding around a focused chip in millimeters (default: 2)",
+    )
     .option("--disable-parts-engine", "Disable the parts engine")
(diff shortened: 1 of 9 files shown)


seveibar:
CC @0hmX @Abse2001 @ShiboSoftwareDev @AnasSarkiz   maybe you would find this interesting, allows `tsci snapshot --chip U1` to take snapshots of a particular chip on a pcb, combined with `--chip-padding` you could use this to investigate issues with ai and have it examine particular regions of the board

PR #3891, opened 2026-07-28
refactor: implement smart entrypoint fallback in getEntrypoint across all CLI commands

KrishnaX12:
## Terminal screenshot : ( Before Vs After) 
area_2026-07-29_00-36-29

## Verification & Testing
- **Unit Testing:** Added full fallback discovery test coverage in `tests/get-entrypoint.test.ts`.
- **Heuristic Ranking Validation:** Added multi-candidate test proving that board files (`my-circuit.circuit.tsx`) are selected over alphabetically earlier helper modules (`a-helper.ts`, `utils.ts`).
- **E2E Publish Verification:** Created integration test in `tests/cli/push/push15-custom-entrypoint-fallback.test.ts` verifying seamless publishing when standard filenames are absent.
- **CI Readiness:** Verified clean execution across all unit and integration testing suites!

### DIFF
--- a/lib/shared/get-entrypoint.ts
+++ b/lib/shared/get-entrypoint.ts
@@ -2,6 +2,7 @@ import * as fs from "node:fs"
 import * as path from "node:path"
 import { loadRuntimeProjectConfig } from "lib/project-config"
 import kleur from "kleur"
+import { findBoardFiles } from "./find-board-files"
 
 type EntrypointOptions = {
   filePath?: string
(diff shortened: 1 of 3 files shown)


seveibar (changes requested):
too hacky

KrishnaX12 (commented):
@seveibar valid lol. I nuked all the weird scoring logic and literally just wrapped our built-in findBoardFiles helper instead. way cleaner now!

seveibar (changes requested):
there is already logic for the default entrypoint, it is a really bad idea to introduce one off logic for this, make sure you're using the latest models b/c this is a strange suggestion

PR #3983, opened 2026-08-02
fix: bridge require("@tscircuit/core") for registry components published before core went ESM-only

### LINKED ISSUE #3982: tsci build: registry components whose CJS build requires @tscircuit/core fail — core is ESM-only (repro of stale #1657)
### Description

In a fresh `tsci init` project, adding some registry components makes `tsci build` fail with:

```
error: Cannot find module '@tscircuit/core' from '.../node_modules/@tsci/Anshgrover23.AMS1117_3_3/index.cjs'
Fatal error [circuit_generation_failed]: ResolveMessage: Cannot find module '@tscircuit/core'
```

### Steps to reproduce

1. `tsci init my-board` (tscircuit 0.0.2205, cli 0.1.1804, bun 1.3.14)
2. `tsci add Anshgrover23/AMS1117_3_3`
3. Use it in `index.circuit.tsx` and run `tsci build`

### Root cause (as far as I can tell)

The registry's CJS build of that component starts with:

```js
var core = require('@tscircuit/core');
```

but `@tscircuit/core` (0.0.1581 in a fresh project) is now ESM-only — `"type": "module"` and an exports map with only an `import` condition:

```json
"exports": { ".": { "import": "./dist/index.js", "types": "./dist/index.d.ts" } }
```

so the `require()` can never resolve. Components with newer registry builds (e.g. `@tsci/seveibar.smd-usb-c`, `@tsci/seveibar.WS2812B_2020`) don't reference `@tscircuit/core` from their CJS output and work fine in the same project — the breakage seems limited to components whose published build predates core going ESM-only (this one imports `createUseComponent` from core).

#1657 reported the same symptom in April; it was read as a not-installed problem and auto-closed by the stale bot, but it still reproduces today with the package above.

### Why it hurts

You only find out at build time, with a module-resolution error that points at a package you installed exactly the way the site suggests. There's no way for a user to tell "this component's published build is stale" from "my project is broken" — I lost a while re-checking my own setup before diffing the working vs failing packages' `index.cjs`. Anything that either rebuilds stale registry artifacts, adds a `require` condition/CJS entry to core, or fails `tsci add` early with "this component needs a rebuild" would fix the experience.

Happy to provide the full project if useful — my workaround was converting the part locally with the easyeda library and importing the generated `.tsx` instead of the registry package.

Hero988:
Fixes #3982 (same failure as #1657, which was closed stale but still reproduces)

### Problem

In a fresh `tsci init` project, `tsci build` fatals when the circuit imports a registry component whose published CJS build starts with `require("@tscircuit/core")`:

```
error: Cannot find module '@tscircuit/core' from '.../node_modules/@tsci/Anshgrover23.AMS1117_3_3/index.cjs'
Fatal error [circuit_generation_failed]: ResolveMessage: Cannot find module '@tscircuit/core'
```

`@tscircuit/core` is ESM-only — its exports map declares only an `import` condition — so that `require()` can never resolve. Every component whose registry build predates core going ESM-only is unusable from `tsci build`, and the error points users at their own setup rather than the stale published artifact.

### Fix

`generateCircuitJson` now registers a small Bun plugin (same registration pattern as `register-static-asset-loaders`) before importing the user's circuit file:

- **Gated:** it first tries `require.resolve("@tscircuit/core")` from the project. If that succeeds (e.g. core ships a `require` condition in the future), the plugin is never registered and nothing changes.
- **Bridge:** when the require is broken, `build.module("@tscircuit/core", ...)` serves the project's ESM build of core to both `require()` and `import` consumers — Bun can require ESM synchronously — so stale CJS registry builds and current ESM code share one core instance.

No behavior change for projects that build fine today: the bridge only registers when resolution is already fatally broken. Under the tsx/Node fallback runtime the function is a no-op (`typeof Bun` guard).

### Test

`tests/cli/build/build-stale-cjs-registry-component.test.ts` reproduces the report: a project with the real ESM-only core symlinked in, plus a minimal `@tsci` package whose CJS entry `require`s core at load time. Without the fix the build fatals with `Cannot find module '@tscircuit/core'`; with it the build exits 0 and emits circuit JSON.

Verified locally: new test passes, `bunx tsc --noEmit` clean, `biome format` clean, `build-transpile` + `build-output-flags` suites still pass (19/19).

### DIFF
--- a/lib/shared/generate-circuit-json.tsx
+++ b/lib/shared/generate-circuit-json.tsx
@@ -10,6 +10,7 @@ import {
   type AutorouterDiagnosticsOptions,
 } from "./autorouter-diagnostics"
 import { importFromUserLand } from "./importFromUserLand"
+import { registerEsmOnlyCoreRequireBridge } from "./register-esm-only-core-require-bridge"
 import { registerStaticAssetLoaders } from "./register-static-asset-loaders"
 import {
   addSourceFilesystemHash,
@@ -95,6 +96,7 @@ export async function generateCircuitJson({
     ? filePath
     : path.resolve(process.cwd(), filePath)
   const projectDir = path.dirname(absoluteFilePath)
+  registerEsmOnlyCoreRequireBridge(projectDir)
   const resolvedOutputDir = outputDir ?? projectDir
 
   // Get the relative path to the component from the project directory
--- a/lib/shared/register-esm-only-core-require-bridge.ts
+++ b/lib/shared/register-esm-only-core-require-bridge.ts
@@ -1,0 +1,78 @@
+import fs from "node:fs"
+import { createRequire } from "node:module"
+import path from "node:path"
+
+let registered = false
+
+const findCorePackageDir = (startDir: string): string | null => {
+  let dir = startDir
+  while (true) {
+    const candidate = path.join(dir, "node_modules", "@tscircuit", "core")
+    if (fs.existsSync(path.join(candidate, "package.json"))) return candidate
+    const parent = path.dirname(dir)
+    if (parent === dir) return null
+    dir = parent
+  }
+}
+
+const getCoreEsmEntry = (corePackageDir: string): string | null => {
+  const packageJsonPath = path.join(corePackageDir, "package.json")
+  if (!fs.existsSync(packageJsonPath)) return null
+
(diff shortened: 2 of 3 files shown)


seveibar:
no need to support this

PR #4042, opened 2026-08-05
Add source diagnostic checks

0hmX:
## Summary

- add the check source subcommand using categorizeErrorOrWarning
- print only source diagnostics from check source
- make plain check print every detected error and warning without category filtering
- retain source and unknown diagnostics even when existing category ignore flags are used
- share the existing Errors and Warnings output formatting across general, source, and netlist checks
- remove the obsolete differential-pair metadata implementation and dependency bump

## Diagnostic behavior

source_property_ignored_warning is visible through both plain check and check source. Plain check is exhaustive and includes source, netlist, placement, routing, pin-specification, and unknown diagnostic types. Category-specific commands remain filtered.

No property-name matching, message parsing, Circuit JSON metadata, schema changes, or producer changes are involved.

## Validation

- source command and exhaustive general-check tests passed
- existing netlist, placement, pin-specification, build-filter, and diagnostic-normalization tests passed
- production build passed against the checked-out circuit-json-util change
- the broader check suite passed except for the pre-existing platform-dependent check-shorts bitmap snapshot mismatch
- git diff --check passed

## Dependency order

This draft depends on tscircuit/circuit-json-util#113 being merged and released. After that release, update the direct util version and regenerate bun.lock before marking this PR ready.


seveibar (changes requested):
Duuude, wrong repo

0hmX:
Closing in favor of a fresh, minimal PR based directly on current main. The replacement removes the old metadata-era history and comments.

PR #4150, opened 2026-08-10
Use packaged check-shorts directly

Abse2001:
## Summary

- remove runtime dynamic loading of `@tscircuit/check-shorts`
- statically import the checker functions used by `tsci check shorts`
- update the packaged checker from 0.0.12 to 0.0.19
- refresh the checker artifact snapshots for 0.0.19
- compare PNG snapshots by decoded pixels so compression differences do not make CI platform-dependent

## Review addressed

The requested dynamic-loading implementation has been removed entirely. The CLI no longer fetches package metadata, downloads or installs a tarball, maintains a runtime cache, imports a file URL, or falls back between online and packaged modules. It now uses the packaged dependency directly.

## Validation

- `bun test tests/cli/check` — 16 passed
- `bun test tests/cli/check/check-shorts.test.ts` — 3 passed
- `bunx tsc --noEmit`
- `bun run format:check`
- `bun run build`
- `git diff --check`


seveibar (changes requested):
Crazy, just remove the dynamic loading altogether

PR #4151, opened 2026-08-10
refactor: prevent command injection

ArnavK-09:
Screenshot From 2026-08-10 18-29-00


seveibar (changes requested):
lots of changes, i don't think this is a real security issue? How could it be exploited

PR #4177, opened 2026-08-11
fix: check shorts on inner copper layers

ShiboSoftwareDev:
## Fix

This is the child fix for #4178. `--layer all` now reads `pcb_board.num_layers`, takes the required layers from the explicit supported list `inner1` through `inner8`, and calls `@tscircuit/check-shorts` once per physical copper layer. The checker API itself already supports inner layers; the missing orchestration was in CLI.

## Snapshot diff

The parent repro shows two crossing `inner1` traces with no marker. This child updates the same `check-shorts-inner-layer-pcb.snap.svg` with the detected-short marker and adds the bitmap diagnostic snapshot.

detected inner-layer short

## Validation

- `bun test tests/cli/check/check-shorts.test.ts` — 6 pass
- `bunx tsc --noEmit`
- `bun run build`
- `bun run format:check`
- before/after SVG snapshots rendered and visually inspected

Stack: #4178 → this PR.


ShiboSoftwareDev:
Superseded by #4181, recreated from a branch owned by the upstream tscircuit repository. Branch chain: #4180 → #4181.

PR #4234, opened 2026-08-13
update fab packages

mohan-bee:


techmannih (approved):

techmannih:
I think it should update automatically in the CLI.

PR #4270, opened 2026-08-14
Preserve QFN thermal vias during CLI footprint imports

techmannih:


scripts/bun-build.ts
```diff
@@ -35,10 +35,18 @@ if (!success) {
   process.exit(1)
 }
 
+const manifoldWasmSourcePath = Bun.resolveSync(
+  "@tscircuit/manifold-2d/manifold.wasm",
+  import.meta.dir,
+)
+const manifoldWasm = Bun.file(manifoldWasmSourcePath)
+await Bun.write("./dist/cli/manifold.wasm", manifoldWasm)
+
```
techmannih, line 44: `circuit-json-to-footprinter@0.0.46` introduces `@tscircuit/manifold-2d` for footprint geometry operations.

The CLI bundles this dependency into `dist/cli/main.js`. In Node.js, Manifold loads `manifold.wasm` relative to that bundled JS file, but Bun does not automatically copy the WASM asset into `dist`.

As a result, the globally installed CLI crashed during startup because it expected:

`dist/cli/manifold.wasm`

Copying the package-exported WASM file next to `main.js` preserves Manifold's runtime loading contract. Only `dist/cli` needs the asset because the other generated bundles do not include the Manifold loader.

techmannih:
sctive

imrishabh18 (commented):
I am not sure, get a review from Seve

scripts/bun-build.ts
seveibar: i don't think this is necessary

seveibar (changes requested):
looks like some crazy manifold hack was introduced, afaik you don't need this hack

techmannih:
Yes but the built CLI demonstrably crashes without it

PR #4353, opened 2026-08-19
fix: allow selecting inner layers in check shorts

AnasSarkiz:
## Summary

- use `circuit-json`'s `LayerRef`, `all_layers`, and `layer_string` as the canonical layer source
- allow `tsci check shorts --layer` to select inner copper layers individually
- validate selected layers against the physical layers derived from `pcb_board.num_layers`
- preserve `--layer all`, including top-only behavior for one-layer boards
- add regression coverage for valid, unavailable, invalid, and one-layer selections

## Root cause

The shorts checker already accepts `LayerRef` values, but the CLI parser was hard-coded to `top`, `bottom`, and `all`. This caused `--layer inner1` to fail before the circuit was loaded or analyzed.

## Impact

Users can request layer-specific short evidence for canonical inner layers that exist on the board. Syntactically valid layers that are not physically present now fail with an available-layers message.

## Validation

- `bun test tests/cli/check/check-shorts.test.ts` — 8 pass
- `bunx tsc --noEmit`
- `bun run format:check`
- `bun run build`

### DIFF
--- a/cli/check/shorts/get-check-short-layers.ts
+++ b/cli/check/shorts/get-check-short-layers.ts
@@ -1,10 +1,19 @@
-import { all_layers, type AnyCircuitElement, type LayerRef } from "circuit-json"
+import type { AnyCircuitElement, LayerRef } from "circuit-json"
 
 export type CheckShortsLayerOption = LayerRef | "all"
 
-const innerCopperLayers = all_layers.filter((layer) =>
-  layer.startsWith("inner"),
-)
+type InnerLayerRef = Exclude<LayerRef, "top" | "bottom">
+
+const innerCopperLayers = [
+  "inner1",
+  "inner2",
+  "inner3",
+  "inner4",
+  "inner5",
+  "inner6",
+  "inner7",
+  "inner8",
+] satisfies InnerLayerRef[]
 
 export const getCheckShortLayers = ({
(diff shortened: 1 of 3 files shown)


cli/check/shorts/get-check-short-layers.ts
```diff
@@ -1,17 +1,10 @@
...
-  "inner2",
-  "inner3",
-  "inner4",
-  "inner5",
-  "inner6",
-  "inner7",
-  "inner8",
-] satisfies LayerRef[]
+const innerCopperLayers = all_layers.filter((layer) =>
+  layer.startsWith("inner"),
```
ShiboSoftwareDev: i think it's better to explicitly name the layers so if the type is updated you have to update it here or typescript would throw, otherwise the regressions/bugs will be hidden

AnasSarkiz: Addressed in 06ca0ebd. I restored the explicit inner1–inner8 list and constrained its entries with InnerLayerRef = Exclude<LayerRef, "top" | "bottom">, so the supported layers remain checked against circuit-json without relying on prefix matching. The focused tests, typecheck, formatting, and build all pass.

ShiboSoftwareDev (approved):

PR #4453, opened 2026-08-24
Warn when importing an out-of-stock JLCPCB part

techmannih:
Screenshot 2026-08-24 at 21 12 37

## Summary

- carry JLCPCB stock information through the `tsci import` flow
- show a warning when the selected part has zero stock
- continue importing the component instead of failing

## Testing

- added a focused test for the out-of-stock warning
- `bun test tests/cli/import/warn-if-jlcpcb-part-is-out-of-stock.test.ts`
- Biome checks pass

## Note

The full build is currently blocked by an unrelated existing `circuit-json-to-gerber` export error.

### DIFF
--- a/cli/import/register.ts
+++ b/cli/import/register.ts
@@ -38,6 +38,7 @@ export const registerImport = (program: Command) => {
         const hasFilters = opts.jlcpcb || opts.lcsc || opts.tscircuit
         const searchJlc = opts.jlcpcb || opts.lcsc || !hasFilters
         const searchTscircuit = opts.tscircuit
+        let didJlcSearchSucceed = false
         const ky = getRegistryApiKy()
         const spinner = ora({
           text: "Searching...",
@@ -55,6 +56,7 @@ export const registerImport = (program: Command) => {
           package: string
           description: string
           price: number
+          stock: number
         }> = []
 
         if (searchTscircuit) {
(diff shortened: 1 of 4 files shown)


cli/import/register.ts
```diff
@@ -173,6 +185,7 @@ export const registerImport = (program: Command) => {
           await importJlcpcbPart({
             download: opts.download,
             partNumber: `C${String(choice.part)}`,
+            stock: choice.stock,
             useExactFootprint: opts.useExactFootprint,
```
imrishabh18, line 197: wrong pattern, why would you pass the stock number to `importJlcpcbPart`. Doesn't make sense

imrishabh18 (changes requested):

cli/import/register.ts
techmannih: Fixed. I moved the stock warning to the command orchestration layer, before calling importJlcpcbPart. The importer no longer receives or handles stock metadata.

cli/import/register.ts
```diff
@@ -97,6 +105,10 @@ export const registerImport = (program: Command) => {
             ? parseDirectLcscPartNumber(query)
             : null
           if (directLcscPartNumber) {
+            warnIfJlcpcbPartIsOutOfStock({
+              partNumber: directLcscPartNumber,
+              stock: didJlcSearchSucceed ? 0 : undefined,
+            })
```
imrishabh18, line 111: You are using some weird pattern, use AI for this it should be simple. Why would you send `0` on success and otherwise `undefined`

imrishabh18 (changes requested):

Issue #120, opened 2025-03-23
Implement the Import Dialog for JLCPCB Components


seveibar:
When someone imports a JLCPCB component, it should use [easyeda-converter](https://github.com/tscircuit/easyeda-converter) to drop the generated file into the `imports/*` directory, e.g. `imports/PICO.tsx`

/bounty $20

seveibar:
CC @ArnavK-09

ArnavK-09:
/attempt #120


<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/ArnavK-09) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @ArnavK-09 | 41 tscircuit bounties | <div align="center">TypeScript, JavaScript, <br />HTML & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm8lv5e7o0002l703e5rq2zgn/cancel-attempt) |

PR #4480, opened 2026-08-25
fix: transpile explicit TypeScript extension imports

seveibar:
## Summary

- preserve explicit `.ts`/`.tsx` imports during library transpilation by enabling TypeScript 5.7+ relative-extension rewriting
- retain the existing fallback for older TypeScript versions
- add a regression test covering a user tsconfig with `allowImportingTsExtensions`

## Cause

The transpiler unconditionally overrode `allowImportingTsExtensions` to `false` because plain TypeScript emit cannot use it. Projects such as `tscircuit/ti` intentionally import `.tsx` files and therefore produced TS5097 for every import during both ESM and CJS bundling. TypeScript 5.7 introduced `rewriteRelativeImportExtensions`, which makes this combination emit-safe.

Reproduction: https://tscircuit.com/tscircuit/ti/releases/1a2d8fe2-3a34-468a-9557-70d7102d837b

## Validation

- `bun test tests/cli/transpile/transpile.test.ts`
- `bun test tests/cli/build/build-transpile.test.ts`
- `bun run format:check`
- `bun run build`
- `bun ../cli/cli/main.ts build --ci --concurrency 4` against tscircuit/ti PR #120: 118/118 circuits passed, followed by ESM, CJS, and declaration generation without TS5097
- the generated ESM, CJS, and declaration artifacts for that PR are byte-identical before and after this change

### DIFF
--- a/cli/build/transpile/index.ts
+++ b/cli/build/transpile/index.ts
@@ -7,11 +7,26 @@ import commonjs from "@rollup/plugin-commonjs"
 import json from "@rollup/plugin-json"
 import dts from "rollup-plugin-dts"
 import kleur from "kleur"
+import ts from "typescript"
 
 import {
   createStaticAssetPlugin,
   STATIC_ASSET_EXTENSIONS,
 } from "./static-asset-plugin"
+
+const [typescriptMajor, typescriptMinor] = ts.versionMajorMinor
+  .split(".")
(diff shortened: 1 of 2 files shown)


seveibar:
CC @0hmX this probably fixes your issues

PR #4666, opened 2026-09-09
Add explicit install and update commands for tscircuit AI skills

rushabhcodes:
An existing tscircuit skill installation never receives newer guidance: `tsci init` checks for `SKILL.md` and skips directories that already contain it. Upgrading the CLI does not refresh those files, even though the installer downloads from the current `tscircuit/skill` repository.

This surfaced during a SparkFun INA237 board conversion. The installed skills lacked the newer guidance to prefer standard footprint strings such as `footprint="0603"`, and the AI created a manual `Capacitor0603Footprint` instead. Updating the upstream skill alone cannot help users whose installed copies remain unchanged. This PR provides an explicit way to refresh that guidance.

The new command supports:

- `tsci setup skills`: install missing project copies while preserving existing installations.
- `tsci setup skills --update`: refresh project installations from `tscircuit/skill`.
- `tsci setup skills --global --update`: refresh home-directory installations.

Updates cover `.claude/skills/tscircuit` and `.agents/skills/tscircuit`, plus an existing legacy `.codex/skills/tscircuit` copy so it cannot retain stale instructions. The existing init flow remains install-only and now prints an update hint when skills are already installed.

The installer downloads the complete skill once, validates that `SKILL.md` exists, and prepares all replacements before modifying installed copies. It preserves previous files, including local edits, under `.tscircuit/skill-updates/<update-id>/backups/` and prints that location. Local edits are backed up rather than merged. Replacement failures trigger rollback; if rollback also fails, the original backups are retained and the error identifies their location. The explicit command exits nonzero on failure. Downloads also preserve binary assets instead of decoding them as text.

Validation:

- `bun test tests/cli/setup`: 12 passing tests, 67 assertions covering installation, update, legacy copies, binary assets, download failures, rollback, backup retention, symlinks, and CLI options/error handling.
- `bun run build`: passed.
- Biome checks on changed TypeScript files and `git diff --check`: passed.
- Ran the real global updater and verified that all three installed copies have matching `SKILL.md` and `FOOTPRINTS.md` files, the standard-footprint guidance is present, and previous copies are backed up.
- `bunx tsc --noEmit`: fails on `CircuitJson` connector-standard type mismatches in unchanged `lib/shared/export-snippet.ts` and `tests/cli/export/export-component-box-3mf.test.ts`; no diagnostics in the changed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Commands:
                                project
   snapshot [options] [path]    Generate schematic and PCB snapshots (add --3d
                                for 3d preview)
-  setup                        Setup utilities like GitHub Actions
+  setup                        Setup utilities like GitHub Actions and AI skills
   install [packageSpec]        Install project dependencies, or install a
                                specific package (e.g., tsci install
                                https://github.com/espressif/kicad-libraries)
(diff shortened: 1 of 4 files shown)


imrishabh18 (approved):
Get a review from Seve as well

PR #4710, opened 2026-09-11
fix: bind converted KiCad reference text to the instance name

### LINKED ISSUE #4708: tsci convert hardcodes KiCad reference placeholder REF** in reusable TSX footprints
## Problem

Converting a KiCad footprint to a reusable TSX component copies its reference placeholder into a literal `text="REF**"`. Instantiating that component as U1 and U2 prints `REF**` twice instead of the instance reference designators.

This is about preserving **reference-field semantics**, not automatically changing text sizes, rearranging arbitrary footprint text, or adding labels to every custom footprint.

## Reproduction

Tested on Linux x64 with Bun 1.4.2, `@tscircuit/cli@0.1.2057`, `tscircuit@0.0.2516` (resolving `@tscircuit/core@0.0.1892`), and TypeScript 5.9.3. These reproductions use only public tscircuit packages and synthetic components.

In an empty directory:

```sh
bun init -y
bun add --exact @tscircuit/cli@0.1.2057 tscircuit@0.0.2516 typescript@5.9.3
```

Use this `tsconfig.json`:

```json
{"compilerOptions":{"jsx":"react-jsx","moduleResolution":"Bundler","module":"ESNext","target":"ESNext","skipLibCheck":true}}
```

`two-pad.kicad_mod`:

```scheme
(footprint "TwoPad"
 (version 20240108)
 (generator "pcbnew")
 (layer "F.Cu")
 (attr smd)
 (fp_text reference "REF**" (at 0 -2) (layer "F.SilkS") (effects (font (size 1 1) (thickness 0.15))))
 (pad "1" smd rect (at -0.8 0) (size 0.9 1) (layers "F.Cu" "F.Paste" "F.Mask"))
 (pad "2" smd rect (at 0.8 0) (size 0.9 1) (layers "F.Cu" "F.Paste" "F.Mask"))
)
```

```sh
bunx tsci convert two-pad.kicad_mod --name TwoPad --output TwoPad.tsx
```

The generated component includes:

```tsx
<silkscreentext pcbX={0} pcbY={2} anchorAlignment="center"
  fontSize={1} font="tscircuit2024" layer="top" text="REF**" />
```

`converted-labels.tsx`:

```tsx
import { TwoPad } from "./TwoPad"
export default () => (<board width={12} height={8}>
 <TwoPad name="U1" pcbX={-3} schX={-2} />
 <TwoPad name="U2" pcbX={3} schX={2} />
</board>)
```

```sh
bunx tsci build converted-labels.tsx
```

In `dist/converted-labels/circuit.json`, `pcb_silkscreen_text` values are **["REF**", "REF**"]**, while source components are named U1 and U2. The build succeeds.

## Expected

The KiCad `fp_text reference` field should become an instance-aware reference label in generated TSX (for example, a supported reference placeholder or a binding to the component's `name`). Preserve its placement and styling, and keep ordinary user text literal.

## Workaround

Change the generated reference text to `text={props.name ?? "REF**"}`. This is a local generated-code workaround, not a proposed universal fallback policy.

## Scope / related reports

The CLI currently delegates this path to `convertCircuitJsonToTscircuit` after KiCad conversion, so the eventual fix may belong in that converter or in preservation of text-role metadata upstream of it. Filing here for the user-facing `tsci convert` reproduction.

- #1708 concerns the opposite conversion direction (mapping names when exporting to KiCad).
- tscircuit/core#2875 concerns generating labels absent from inline footprints. Here the label exists but has lost its reference semantics.

billythompsons:
Fixes #4708.

## Problem
`tsci convert` copies the KiCad `fp_text reference` placeholder into a literal `text="REF**"`, so every placed instance of the converted footprint prints "REF**" instead of its reference designator.

## Fix
After `convertCircuitJsonToTscircuit` runs, the KiCad reference placeholder (`text="REF**"` on a `silkscreentext`) is rebound to the component instance: `text={props.name ?? "REF**"}`. Ordinary user text stays literal, and the rewrite only applies when the generated component has `props` in scope.

Kept this in the CLI's convert path since the placeholder is a KiCad artifact known at conversion time; happy to move it into `circuit-json-to-tscircuit` if you'd rather handle it there.

## Tests
- `convert-kicad-mod-reference-text.test.ts`: converts the issue's two-pad `.kicad_mod` plus an `fp_text user` control - asserts the reference placeholder is rebound and user text stays literal.
- `convert-kicad-mod-reference-e2e.test.ts`: builds a board with two instances (U1, U2) and asserts the emitted `pcb_silkscreen_text` values are exactly `["U1", "U2"]` (the issue's failing scenario).

Full `tests/cli/convert` suite passes.

### DIFF
--- a/cli/convert/convert-kicad-footprint-to-tsx.ts
+++ b/cli/convert/convert-kicad-footprint-to-tsx.ts
@@ -15,11 +15,25 @@ export const convertKicadFootprintToTsx = async ({
 }) => {
   const circuitJson = await convertKicadFootprintToCircuitJson(inputPath)
   const componentName = name ?? path.basename(inputPath, ".kicad_mod")
-  const tsx = convertCircuitJsonToTscircuit(circuitJson, { componentName })
+  const rawTsx = convertCircuitJsonToTscircuit(circuitJson, { componentName })
+  const tsx = bindKicadReferenceText(rawTsx)
   const outputPath = output
     ? path.resolve(output)
     : path.join(path.dirname(inputPath), `${componentName}.tsx`)
 
   await fs.writeFile(outputPath, tsx)
   console.log(kleur.green(`Converted ${outputPath}`))
+}
+
+// KiCad footprints use the literal "REF**" as the placeholder text of the
+// fp_text reference field. Bind that label to the component instance name so
+// each placed instance prints its own reference designator instead of
+// "REF**". Ordinary user text stays literal.
+const bindKicadReferenceText = (tsx: string) => {
+  // Only rewrite when the generated component has `props` in scope
+  if (!tsx.includes("(props:")) return tsx
+  return tsx.replaceAll(
+    /(<silkscreentext\b[^>]*?)text="REF\*\*"/g,
+    '$1text={props.name ?? "REF**"}',
+  )
 }
--- a/tests/cli/convert/convert-kicad-mod-reference-e2e.test.ts
+++ b/tests/cli/convert/convert-kicad-mod-reference-e2e.test.ts
@@ -1,0 +1,43 @@
+import { getCliTestFixture } from "../../fixtures/get-cli-test-fixture"
+import { test, expect } from "bun:test"
+import { writeFile, readFile } from "node:fs/promises"
+import path from "node:path"
+
+const kicadMod = `(footprint "TwoPad"
+ (version 20240108)
+ (generator "pcbnew")
+ (layer "F.Cu")
+ (attr smd)
+ (fp_text reference "REF**" (at 0 -2) (layer "F.SilkS") (effects (font (size 1 1) (thickness 0.15))))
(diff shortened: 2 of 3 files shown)


billythompsons:
Still active and ready for review. All six checks are green; the e2e test builds two instances of the converted footprint and verifies the silkscreen refs resolve to U1/U2 instead of the REF** placeholder.

billythompsons:
Status update: the focused regression fix for the requested CLI behavior remains green in CI and ready for review. I can rebase or make any requested changes promptly.

billythompsons:
Status update: the focused CLI regression fix remains green in CI and ready for review. I can rebase or make any requested changes promptly.

billythompsons:
Rechecked against current main (fcbf759): the branch still merges cleanly and #4708 still reproduces on main - `tsci convert` on a KiCad footprint emits the literal `REF**` silkscreen text, so every instance renders the placeholder instead of U1/U2. This PR binds the converted reference text to the instance name, and the e2e test covers the two-instance case. Glad to rebase or adjust if you'd prefer a different approach.

PR #4810, opened 2026-09-18
test: verify routing-disabled preserves connected-board placement

devamkakoty:
## Summary

I added regression coverage for the placement-preview portion of
tscircuit/tscircuit#4948. This does not change production routing behavior or
claim to complete that umbrella issue.

The existing test only builds an unconnected resistor. The new test uses two
separated resistors with an explicit pin-to-pin connection:

- With routing enabled, a local custom-router callback records the number of
  connections and deliberately throws a sentinel error. The test requires a
  real invocation with a positive connection count, so an ignored callback
  cannot produce a vacuous pass.
- A fresh fixture builds identical source with `--routing-disabled`. It must
  succeed without the router marker, preserve the source, and emit Circuit JSON
  retaining board dimensions/layers, component identities and placements,
  source-to-PCB port mappings, and the requested logical connection.
- The disabled build must emit no PCB traces or vias. Neither fixture supplies
  a route cache; the parts engine is disabled and the custom router is local.

This is coverage for one connected-board case, not a proof about manual copper,
pours, all routing providers, diagnostics, or 3D preview.

## Validation

Run on Windows with Bun 1.4.2 against the committed lockfile:

- `bun test tests/cli/build/build-routing-disabled.test.ts --timeout 120000`
  — **2 pass, 0 fail, 51 assertions**.
- `bun run format:check` — passed across 457 files.
- `bun run build` — passed.
- Local installed TypeScript: `bun node_modules/typescript/bin/tsc --noEmit`
  — passed.
- `git diff --check` — passed.

The full test suite was not run locally.

AI assistance was used for implementation and review. The checks above were
actually executed; this is not a report of a newly discovered production bug.


devamkakoty:
I'd welcome a review of this as a standalone regression-test contribution. All 10 upstream checks were passing in my latest check, and the submitted head is unchanged.

The scope remains the connected-board `--routing-disabled` case described above, not completion of the broader placement-preview issue. The executed validation and AI-assistance disclosure remain in the PR description.

Is this useful to merge independently, or would you prefer a specific adjustment to the test coverage?

PR #4830, opened 2026-09-19
fix: resolve CLI version dynamically from package.json without synthetic patch increment

### LINKED ISSUE #4628: Update prompt loops on every run (0.1.2020 → 0.1.2021): bundled version is stale because build runs before `pver release`
## Summary

Every run of `tsci` (e.g. `tsci init`) on the current latest release shows:

```
A new version of tsci is available (0.1.2020 → 0.1.2021).
Would you like to update now?
```

Accepting runs `npm install -g tscircuit@latest` (or the bun equivalent), which reinstalls the exact same version, and the prompt appears again on the next run. There is no way to satisfy the check.

Same symptom as the old #155, but with a different and reproducible root cause this time.

## Environment

- `tscircuit@0.0.2465` (npm `latest`), which depends on `@tscircuit/cli@^0.1.2021`
- `@tscircuit/cli@0.1.2021` (npm `latest`)
- Linux, Node v22.23.0 and Bun 1.2.14 — reproduces under both runtimes

## Root cause

`currentCliVersion()` in `lib/shared/check-for-cli-update.ts` is:

```ts
export const currentCliVersion = () =>
  program?.version() ?? semver.inc(pkgVersion, "patch") ?? pkgVersion
```

where `pkgVersion` is imported from `package.json` and therefore **inlined into `dist/cli/main.js` at build time**. The `+1 patch` assumes the published version is always exactly one patch above the committed `package.json` version.

That assumption breaks because `.github/workflows/bun-pver-release.yml` runs `bun run build` **before** `pver release` bumps the version, and the committed `package.json` has fallen behind:

| Source | Version |
|---|---|
| `package.json` in `main` (last bump PR merged: #4510, 2026-08-27) | 0.1.2019 |
| `var version = "…"` inlined in the published `dist/cli/main.js` | 0.1.2019 |
| What the CLI reports for itself (`0.1.2019` + 1 patch) | 0.1.2020 |
| `package.json` shipped on disk in `@tscircuit/cli@0.1.2021` | 0.1.2021 |
| `https://registry.npmjs.org/@tscircuit/cli/latest` | 0.1.2021 |

So `semver.gt("0.1.2021", "0.1.2020")` is true forever for this build, and `tsci --version --verbose` prints `@tscircuit/cli: 0.1.2020` even though the installed package is 0.1.2021.

Verified locally:

```
$ grep '"version"' node_modules/tscircuit/node_modules/@tscircuit/cli/package.json
  "version": "0.1.2021",
$ grep -n -m1 'var version = ' node_modules/tscircuit/node_modules/@tscircuit/cli/dist/cli/main.js
136147:var version = "0.1.2019";
$ tsci --version --verbose
tscircuit: 0.0.2465
@tscircuit/cli: 0.1.2020
```

## Why the gap widened (and why no fix has shipped)

The version-bump PRs stopped merging after v0.1.2019, and the **Publish to npm** workflow has been failing on every push to `main` since (e.g. run 34154102423 on 2026-09-07). `pver release` starts from `package.json` (0.1.2019), finds tags 0.1.2020 … 0.1.2030 already taken, and aborts:

```
current version: 0.1.2019
candidate next version: 0.1.2020
Next version already exists as a git tag: 0.1.2020. Incrementing version by an increment.
…
Error: Next version already exists as a git tag: 0.1.2030. We tried to incremenet 10 times but the tag still exists. Consider deleting some git tags.
```

Git tags exist up to `v0.1.2030` but npm stops at `0.1.2021`, so nothing has been published since, and every user on `tscircuit@latest` is stuck in the update loop.

## Suggested fixes

1. **Stop guessing the version.** Read `require("@tscircuit/cli/package.json").version` at runtime (the on-disk file is correct: 0.1.2021) instead of inlining `package.json` and adding a patch. Alternatively, run `bun run build` *after* `pver release` bumps `package.json`, so the inlined value is right.
2. **Unblock the release pipeline**: either bump `package.json` in `main` past `0.1.2030` or delete the orphan tags `v0.1.2022`–`v0.1.2030`, and look into why the auto-merge of `version-bumps/**` PRs stopped after #4510.
3. Minor: the check compares against `@tscircuit/cli@latest` but installs `tscircuit@latest`. If `tscircuit` has not yet been republished with the newer CLI, "updating" is a no-op and the prompt persists. Comparing against `tscircuit@latest`'s pinned CLI version would avoid that.

## Workaround for users

```bash
export TSCI_SKIP_CLI_UPDATE=true
```

meanusarcanus:
## Description
Fixes #4628

### Root Cause
Previously, `currentCliVersion()` in `lib/shared/check-for-cli-update.ts` and `getCliVersion()` in `lib/getVersion.ts` relied on an inlined `package.json` version at bundle time and guessed the published version with `semver.inc(pkgVersion, "patch")`.

When `bun run build` runs before release bumps or when committed `package.json` lags behind git tags, the inlined version remains stale in `dist/cli/main.js`. For instance, in `@tscircuit/cli@0.1.2021`, the on-disk `package.json` was `0.1.2021`, but the bundled CLI computed `0.1.2019 + 1 patch = 0.1.2020`. This triggered an infinite update prompt loop on every execution of `tsci`.

### Changes
1. **Dynamic Runtime Resolution (`lib/getVersion.ts`)**:
   - Updates `getCliVersion()` to read the real on-disk `package.json` version dynamically at runtime:
     - Resolves via node module resolution (`@tscircuit/cli/package.json`)
     - Walks up directory tree from current file location to locate the closest `@tscircuit/cli` package root
     - Falls back to `pkg.version` without synthetic patch guessing
2. **Unified Version Helper (`lib/shared/check-for-cli-update.ts`)**:
   - Replaces inlined `pkgVersion` and `semver.inc` with `getCliVersion()`, ensuring the update checker and `tsci --version` report the exact same genuine installed version.
3. **Tests**:
   - Added unit test in `tests/lib/getVersion.test.ts` verifying `getCliVersion()` matches the real package version.
   - Added unit tests in `tests/lib/check-for-cli-update.test.ts` for update checks and version resolution.

@algora-pbc /claim #4628


meanusarcanus:
### 💰 Payout Details
- **PayPal**: `tednadres@yahoo.com`

